PR #1, opened 2021-08-13
Improve node probes checks

PierreBesson:
- add default startup probe (http check on /health every 10s for 5m)
- add readiness check (tcp socket check on the WS port every 10s for 24h)
- add --ws-external as default flag to make the readiness check work in the default config
- allow disabling the readiness check

### DIFF
--- a/charts/node/templates/statefulset.yaml
+++ b/charts/node/templates/statefulset.yaml
@@ -130,16 +130,30 @@ spec:
             - containerPort: 30333
               name: p2p
               protocol: TCP
+          {{- if .Values.node.enableStartupProbe }}
+          # On startup, retry the connection to the /health endpoint every 10s for 5 min before killing the container
+          startupProbe:
+            failureThreshold: 30
+            periodSeconds: 10
+            # Use an exec command as Substate's default configuration only allow local network access to the RPC endpoint
+            exec:
+              command:
+              - curl
+              - "http://127.0.0.1:9933/health"
+          {{- end }}
+          {{- if .Values.node.enableReadinessProbe }}
+          # Continuously retry the connection to the WS endpoint every 10s for 24h until success before marking the container as ready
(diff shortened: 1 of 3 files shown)


charts/node/values.yaml
```diff
@@ -57,9 +57,12 @@ node:
   #chainPath: ""
   #chainDataKubernetesVolumeSnapshot: ""
   #chainDataGcsBucketUrl: ""
+  enableStartupProbe: true
+  enableReadinessProbe: true
   flags:
     - "--prometheus-external"
     - "--rpc-external"
+    - "--ws-external"
     - "--rpc-cors"
```
lovelaced, line 66: if you want to make this part more clear you could just use "--rpc-cors=all" on a single line.

lovelaced (approved):

charts/node/values.yaml
PierreBesson: Thanks I changed it to what you suggested.

charts/node/templates/statefulset.yaml
```diff
@@ -130,16 +130,30 @@ spec:
...
+          {{- if .Values.node.enableStartupProbe }}
+          # On startup, retry the connection to the /health endpoint every 10s for 5 min before killing the container
+          startupProbe:
+            failureThreshold: 30
+            periodSeconds: 10
+            # Use an exec command as Substate's default configuration only allow local network access to the RPC endpoint
+            exec:
+              command:
+              - curl
+              - "http://127.0.0.1:9933/health"
```
gabreal, line 142: what's the difference to using the [httpGet call](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes)?

PierreBesson: In case `--rpc-external` is not enabled on the node the httpGet call will fail as will be blocked by substrate as an external network call. There might be cases where users want to run nodes without this flag enabled, so we do this call from within the container itself where it will never be blocked.

PR #4, opened 2021-08-19
adding substrate-telemetry helm chart

arshamparity:
according to the conversation on  https://github.com/paritytech/substrate-telemetry/pull/368, a new generic helm chart is created for the `minikube` users or everyone that have a basic Kubernetes cluster running without fancy features we have on our clusters.
This helm chart is meant to be a basic installation for telemetry service on every Kubernetes cluster. The readme has some notes for public users who want to run their own telemetry instance.

### DIFF
--- a/charts/substrate-telemetry/values.yaml
+++ b/charts/substrate-telemetry/values.yaml
@@ -1,0 +1,160 @@
+# Default values for substrate-telemtry.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+
+replicaCount:
+  # NOTE: The core service is not scalable at the moment.
+  shard: 1
+  core: 1
+  frontend: 1
+
+image:
+  backend:
+    repository: docker.io/parity/substrate-telemetry-backend
+    pullPolicy: IfNotPresent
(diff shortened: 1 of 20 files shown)


charts/substrate-telemetry/values.yaml
```diff
@@ -0,0 +1,160 @@
...
+        paths:
+          - path: /
+            pathType: ImplementationSpecific
+    tls:
+      - secretName: chart-example-tls
+        hosts:
+          - chart-example.local
+  core:
+    enabled: false
+    className: ""
```
PierreBesson, line 103: Can this section be commented out as those will never be correct values.

PierreBesson (approved):

charts/substrate-telemetry/values.yaml
arshamparity: I've set `enabled: false` so these are not included in the helm template. I left them uncommented as most of the users might want to use ingress for exposing the services.

PierreBesson:
Please add a link to the new chart on the main [README](https://github.com/paritytech/helm-charts/blob/main/README.md).

PR #13, opened 2021-09-13
Set external traffic policy to local for node port services

PierreBesson:
To help with connecting to nodes p2p ports directly without traffic being shifted between nodes by kubernetes.

From [Kubernetes Reference Service API v1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#service-v1-core):

> externalTrafficPolicy denotes if this Service desires to route external traffic to node-local or cluster-wide endpoints. "Local" preserves the client source IP and avoids a second hop for LoadBalancer and Nodeport type services, but risks potentially imbalanced traffic spreading. "Cluster" obscures the client source IP and may cause a second hop to another node, but should have good overall load-spreading.

### DIFF
--- a/charts/node/templates/service.yaml
+++ b/charts/node/templates/service.yaml
@@ -51,6 +51,7 @@ metadata:
   name: {{ $fullname }}-{{ $i }}-p2p
 spec:
   type: NodePort
+  externalTrafficPolicy: Local
   selector:
     {{- $selectorLabels | nindent 4 }}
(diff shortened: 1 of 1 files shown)


arshamparity:
looks great!
There's also a `ServiceInternalTrafficPolicy` which I think should be enabled once it becomes `stable`. It improves the performance by descreasing the internal traffic between the Kubernetes nodes. But it's now part of `alpha`. (just mentioning for the future)
https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/#using-service-internal-traffic-policy

PR #14, opened 2021-09-13
Allow persisting the auto-generated node key

PierreBesson:


charts/node/values.yaml
```diff
@@ -49,6 +49,7 @@ ingress:
 
 node:
   chain: "polkadot"
+  command: "polkadot"
```
arshamparity, line 52: isn't the command always `polkadot`?

PierreBesson: It is for westend, kusama and polkadot but not for other substrate chains and parachains.

PR #16, opened 2021-09-14
Allow setting the node role in the chart options

PierreBesson:


gabreal:
wouldnt' we also want a label set by that option? such like `role: validator` for being able to query the corresponding pods later.

Issue #26, opened 2021-11-16
Add jaeger-agent/grafana-agent sidecar option


PierreBesson:
Add an option to the node helm chart for doing the following:

- Run [jaeger-agent](https://www.jaegertracing.io/docs/1.28/deployment/#agent) or [grafana agent](https://grafana.com/docs/tempo/latest/grafana-agent/) as a sidecar of the node pod to receive traces
- Configure the agent to forward traces to the Jaeger/Tempo collector URL
- Add the `--jaeger-agent 127.0.0.1:$AGENT_PORT` flag to the node

lazam:
Resolving this one since #27 is merged already.

PR #27, opened 2021-11-18
Add jaeger agent as sidecar

lazam:
* Include Jaeger agent as sidecar
* Move substrate sidecar under sidecar value

### DIFF
--- a/charts/node/README.md
+++ b/charts/node/README.md
@@ -65,29 +65,30 @@ helm install kusama-node parity/node --set node.chainDataSnapshotUrl=https://ksm
 | `node.serviceMonitor.namespace`          | Prometheus namespace                                                                                                      | `nil`                          |
 | `node.serviceMonitor.internal`           | Prometheus scrape interval                                                                                                | `nil`                          |
 | `node.serviceMonitor.scrapeTimeout`      | Prometheus scrape timeout                                                                                                 | `nil`                          |
+| `node.tracing.enabled`                   | If true, creates a jaeger agent sidecar                                                                                   | `false`                        |
+| `node.subtrateApi.enabled`               | If true, creates a substrate api sidecar                                                                                  | `false`                        |
 
 ### Other parameters
 
-| Parameter                          | Description                                                                                            | Default                   |
-|------------------------------------|--------------------------------------------------------------------------------------------------------|---------------------------|
-| `image.repository`                       | Node image name                                                                                        | `parity/polkadot`   |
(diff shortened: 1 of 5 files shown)


lazam:
Resolves #26

charts/node/README.md
```diff
@@ -68,18 +68,26 @@ helm install kusama-node parity/node --set node.chainDataSnapshotUrl=https://ksm
...
+| `ingress.enabled`                        | If true, creates an ingress                                                                            | `false`             |
+| `ingress.annotations`                    | Annotations to add to the ingress (key/value pairs)                                                    | `{}`                |
+| `ingress.rules`                          | Set rules on the ingress                                                                               | `[]`                |
+| `ingress.tls`                            | Set TLS configuration on the ingress                                                                   | `[]`                |
+| `podSecurityContext`                     | Set the pod security context for the substrate node container                                          | `{ runAsUser: 1000, runAsGroup: 1000, fsGroup: 1000 }`|
+| `sidecar.jaegerAgent.enabled`            | If true, enabled Jaeger agent as sidecar                                                               | `false`             |
```
PierreBesson, line 86: I would use the following variable name: `node.tracing.enabled` to enable tracing on the node and based on that activate the jaeger agent and use the properties such as `jaegerAgent.collector.url`...

lazam: Addressed this with the latest commit (a5ad3c9df73)

charts/node/templates/statefulset.yaml
```diff
@@ -225,21 +231,55 @@ spec:
           volumeMounts:
             - mountPath: /data
               name: chain-data
-        {{- if .Values.substrateApiSidecar.enabled }}
+        {{- if .Values.node.substrateApi.enabled }}
```
PierreBesson, line 234: should be `substrateApiSidecar` here.

charts/node/values.yaml
```diff
@@ -89,14 +89,41 @@ node:
...
     repository: parity/substrate-api-sidecar
     tag: latest
   env: {}
   resources: {}
 
+# https://github.com/jaegertracing/jaeger
+jaegerAgent:
+  image:
+    repository: jaegertracing/jaeger-agent
+    tag: 1.28.0
```
PierreBesson, line 112: Let's use `latest` here and not set a default version here as we can't always keep it up date. Also, automated helm security analysis tools will assign a bad score to your chart because it has old version of the docker images. However when using the chart, a versioned tag should always be set.

PR #28, opened 2021-11-26
P2P address auto discovery

PierreBesson:
Implement a new way of configuring the node external p2p address exposed on a Kubernetes nodePort service.

- Auto-discover the external port, by querying the kubernetes service to retrieve the attributed node port
- Auto-discover the external ip address from a third party service such as ifconfig.io (optional, shouldn't be needed if Kubernetes nodes have public IPs as it should be autodiscovered by libp2p)


gabreal (approved):
very nice!

Issue #39, opened 2022-01-13
Method for adding custom node-keys


dblane-digicatapult:
In our use case we use the `pallet-node-authorization` and we need to be able to insert the `node-key`.

Currently the only methods for achieving this are:

- Bring up a node with `node.persistGeneratedNodeKey` true and then `kubectl exec` into the container and put our key in `/data/node-key` OR
- Bring up a node with `node.persistGeneratedNodeKey` true and take the generated `node-key` and recreate this in our node-authorization pallet, recreate the chainspec then download it using the `node.customChainspecUrl`

A better solution would be allowing us the option to create our own `node-key` mounted as a read-only `secret` instead of reading it from the RW data volume.

We can provide more detail if requested.

PR #40, opened 2022-01-18
Add custom node-key

dblane-digicatapult:
This PR fulfils #39 

The decision to add using the `-node-key` argument is because the node requires the location of `-node-key-file` to be a RW destination.
```
Error: Service(Network(Read-only file system (os error 30)))
```
If `persistGeneratedNodeKey` is true then we will not try and use a custom node-key, create a secret or try and mount that secret.


PierreBesson (approved):

PierreBesson:
Thank you for your contribution. Sorry for the late review, I will try to keep a closer eye on PRs and issues.

PR #42, opened 2022-02-21
Added a method for inserting keys without revealing seed values

### LINKED ISSUE #38: inject-keys init container displays keys
When inserting keys via the `values.node.keys` method the init container displays them, meaning that anyone who has read access to the `statefulSet` can read them.

It would probably be better to mount these as `secret` and inject them using a file redirect than an echo.

dblane-digicatapult:
This fixes #38 by creating a secret for each key type and then inserting it using cat's instead of echoing.

### DIFF
--- a/charts/node/Chart.yaml
+++ b/charts/node/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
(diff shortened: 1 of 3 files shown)


n3op2 (approved):

PierreBesson (approved):

PierreBesson:
Thanks for your contribution @dblane-digicatapult

PR #46, opened 2022-02-23
Add support for moonriver-based collators

BulatSaif:
Moonriver-based collators use chainspec file names to define internal logic (see: https://github.com/PureStake/moonbeam/issues/1104#issuecomment-996787548 )
In current version of  helm chart chainspec the  file name is hard-coded to `/data/chainspec.json`
This PR adds support for such nodes, by allowing users to change the name of chainspec file. 

The variable `customChainspecPath` should contain the full path to chainspec inside a container. The reason why I moved full path to a variable, not just name because in the future we may add the feature to set chain spec as `configmap` and we should  be able to use this value. (However, is unlikely due to `configmap` size limit of 1MB.)

This change is backward compatible since it sets the default name to the old value.

### DIFF
--- a/charts/node/Chart.yaml
+++ b/charts/node/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
-version: 0.24.1
+version: 0.24.2
 appVersion: "0.0.1"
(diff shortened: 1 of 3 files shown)


PierreBesson:
This makes me think of this code in Cumulus which selects the chainspec depending on the first chars of the chain ID: https://github.com/paritytech/cumulus/blob/2bcd7b535a904b646a8055cfff0b2fe6ac37c1e1/polkadot-parachains/src/command.rs#L53

PR #52, opened 2022-03-02
Include additional custom sidecar

lazam:
To allow users to add their own custom sidecar


charts/node/README.md
```diff
@@ -106,3 +106,4 @@ helm install kusama-node parity/node --set node.chainDataSnapshotUrl=https://ksm
 | `jaegerAgent.ports.samplingPort`   | Port for HTTP sampling strategies                                                                      | `5778`              |
 | `jaegerAgent.collector.url`        | The URL which jaeger agent sends data                                                                  | `nil`               |
 | `jaegerAgent.collector.port   `    | The port which jaeger agent sends data                                                                 | `14250`             |    
+| `extraContainers   `               | Sidecar containers to add to the node                                                                  | `""`                |
```
PierreBesson, line 109: Why not call this `extraSidecarContainers` ?

lazam: I think its redundant to add a sidecar in the name since this helm will deploy node, and any additional containers except initContainers that will be created is expected to be a sidecar. If we plan to add more deployments/statefulset for this chart then I can move it under `.Values.node.extraContainers`

Issue #50, opened 2022-02-28
Using HashiCorp Vault in Helm Charts


kaverikhaneja:
Would like to use HashiCorp Vault in Susbtrate/Polkadot helm charts for secret management and protecting sensitive data, in this case, the node keys, for instance.

PierreBesson:
We have achieved such an integration with other Kubernetes deployments, using the [Vault agent injector](https://www.vaultproject.io/docs/platform/k8s/injector) and the [proper annotations](https://www.vaultproject.io/docs/platform/k8s/injector/annotations). However configuring the Vault itself can be quite complex and is outside the scope of the helm-chart. We should just let users map the KV paths and policy they want to use for each key, leaving authentication concerns to the cluster administrator.

PR #55, opened 2022-03-05
Vault annotations integration

dblane-digicatapult:
Integrates vault using [annotations](https://www.vaultproject.io/docs/platform/k8s/injector/annotations) as per the [vault agent injector](https://www.vaultproject.io/docs/platform/k8s/injector).

I have tested this reasonable extensively however this would definitely benefit from having another eye cast over it and some more rigorous  testing.

Should implement #50

### DIFF
--- a/charts/node/templates/statefulset.yaml
+++ b/charts/node/templates/statefulset.yaml
@@ -241,7 +241,6 @@ spec:
             - -c
             - |
               {{- range $keys := .Values.node.vault.keys }}
-              cat /vault/secrets/{{ .name }}
               {{ $.Values.node.command }} key insert --base-path /data \
               --chain ${CHAIN} \
               --key-type $(awk '/{{ .typeTarget }}/ {print $2}' /vault/secrets/{{ .name }}) \
--- a/charts/node/templates/statefulset.yaml
+++ b/charts/node/templates/statefulset.yaml
@@ -17,9 +17,30 @@ spec:
   serviceName: {{ $fullname }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
+      {{- if or .Values.podAnnotations .Values.node.vault.keys .Values.node.vault.nodeKey }}
       annotations:
+      {{- with .Values.podAnnotations }}
       {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- range $keys := .Values.node.vault.keys }}
+        vault.hashicorp.com/agent-inject-secret-{{ .name }}: {{ .vaultPath | squote }}
+        vault.hashicorp.com/agent-inject-template-{{ .name }}: |
+          {{`{{ with secret "`}}{{ .vaultPath }}{{`" }}{{- range $k, $v := .Data.data }}
+          {{ $k }}: {{ $v }}
+          {{- end }}{{ end }}`}}
+      {{- end }}
+      {{- if or .Values.node.vault.keys .Values.node.vault.nodeKey }}
+      {{- if .Values.node.vault.nodeKey }}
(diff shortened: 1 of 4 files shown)


charts/node/templates/statefulset.yaml
```diff
@@ -17,10 +17,31 @@ spec:
...
       annotations:
+      {{- with .Values.podAnnotations }}
       {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- range $keys := .Values.node.vault.keys }}
+        vault.hashicorp.com/agent-inject-secret-{{ .name }}: {{ .vaultPath | squote }}
+        vault.hashicorp.com/agent-inject-template-{{ .name }}: |
+          {{`{{ with secret "`}}{{ .vaultPath }}{{`" }}{{- range $k, $v := .Data.data }}
+          {{ $k }}: {{ $v }}
+          {{- end }}{{ end }}`}}
```
dblane-digicatapult, line 30: I am not particularly proud of this but it was the easiest way of getting values out of the secrets without having to evaluate a map inside our init container.

charts/node/templates/statefulset.yaml
```diff
@@ -212,6 +233,28 @@ spec:
...
+          command: [ "/bin/sh" ]
+          args:
+            - -c
+            - |
+              {{- range $keys := .Values.node.vault.keys }}
+              {{ $.Values.node.command }} key insert --base-path /data \
+              --chain ${CHAIN} \
+              --key-type $(awk '/{{ .typeTarget }}/ {print $2}' /vault/secrets/{{ .name }}) \
+              --scheme $(awk '/{{ .schemeTarget }}/ {print $2}' /vault/secrets/{{ .name }}) \
+              --suri "$(awk '/{{ .seedTarget }}/ { $1=""; print}' /vault/secrets/{{ .name }} | cut -c 2-)" \
```
dblane-digicatapult, line 248: These `awk` commands could definitely use someone with better bash-fu than myself sanity checking them.

charts/node/templates/statefulset.yaml
```diff
@@ -17,10 +17,31 @@ spec:
...
+        vault.hashicorp.com/agent-inject-template-{{ .name }}: |
+          {{`{{ with secret "`}}{{ .vaultPath }}{{`" }}{{- range $k, $v := .Data.data }}
+          {{ $k }}: {{ $v }}
+          {{- end }}{{ end }}`}}
+      {{- end }}
+      {{- if or .Values.node.vault.keys .Values.node.vault.nodeKey }}
+      {{- if .Values.node.vault.nodeKey }}
+        vault.hashicorp.com/agent-inject-secret-{{ .Values.node.vault.nodeKey.name }}: {{ .Values.node.vault.nodeKey.vaultPath | squote }}
+        vault.hashicorp.com/agent-inject-template-{{ .Values.node.vault.nodeKey.name }}: |
+          {{`{{ with secret "`}}{{ .Values.node.vault.nodeKey.vaultPath }}{{`" }}{{- range $k, $v := .Data.data }}
```
PierreBesson, line 36: I'm not a big fan of this approach to pull all KV into a big pseudo YAML file. I think it might cause issue if the value contains spaces such as when using a seed phrase. I would prefer to save each secret that we use into its own file. I will push a PR with implementation of this method.

charts/node/templates/statefulset.yaml
PierreBesson: If possible, I would prefer to avoid complex awk queries if the secret can be properly saved to a temporary file. It would also help not having to template/parse the same secrets twice (first in the go template then with awk commands).

dblane-digicatapult:
Closing in favour of #61

PR #57, opened 2022-03-07
Include custom target labels for service monitor

lazam:
* Allow the chart to add custom target labels in service monitor


charts/node/values.yaml
```diff
@@ -102,8 +102,11 @@ node:
   serviceMonitor:
     enabled: false
     #namespace: monitoring
-    #interval: 10s
-    # scrapeTimeout: 10s
+    interval: 1m
+    scrapeTimeout: 30s
+    targetLabels:
+      - chain
+      - instance
```
PierreBesson, line 109: Should we really have this `instance` label by default. What does this correspond to ?

lazam: Done!

charts/node/values.yaml
```diff
@@ -102,8 +102,11 @@ node:
   serviceMonitor:
     enabled: false
     #namespace: monitoring
-    #interval: 10s
-    # scrapeTimeout: 10s
+    interval: 1m
+    scrapeTimeout: 30s
+    targetLabels:
+      - chain
```
gabreal, line 108: if i understand correctly this label would conflict with the one from the directly exposed labels from the metrics.

charts/node/values.yaml
```diff
@@ -102,8 +102,11 @@ node:
   serviceMonitor:
     enabled: false
     #namespace: monitoring
-    #interval: 10s
-    # scrapeTimeout: 10s
+    interval: 1m
+    scrapeTimeout: 30s
+    targetLabels:
```
gabreal, line 107: setting this up will be a bit tricky and would probably be good to mention here: one can only target labels that have been added either by the `serviceLabels` or `extraLabels`.

charts/node/values.yaml
gabreal: hmm, i think the reason behind that `instance` label was (though not targetted through) to have a common label for vm and kubernetes based deployments. would that be `node` then in that case?

charts/node/values.yaml
PierreBesson: Oh yes, it's not good to have conflicts, then it would remap the existing `chain` label to `exported_chain` which is really confusing. See [prometheus configuration docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration).

```
# [...] label conflicts are resolved by renaming
# conflicting labels in the scraped data to "exported_<original-label>" (for
# example "exported_instance", "exported_job") and then attaching server-side
# labels.
```

PierreBesson: In this case, it's probably better to only add `node` to the default `targetLabels`

charts/node/values.yaml
PierreBesson: Yes node would be the new label. @lazam `instance` also need to be renamed to `node` here: https://github.com/paritytech/helm-charts/blob/main/charts/node/templates/service.yaml#L33

charts/node/values.yaml
lazam: I removed the chain in the default values, and added `honor_labels` to avoid having conflict when a label is provided by he metrics.

```
If honor_labels is set to "true", label conflicts are resolved by keeping label
values from the scraped data and ignoring the conflicting server-side labels.
```

charts/node/values.yaml
lazam: I added the details in the description for the parameters in  README.md

PierreBesson:
Great it looks good to me. I didn't know that you could set honorLabel directly there. It will help others not have the same kind of confusion we had with those `exported_` labels.

PierreBesson (approved):

PR #61, opened 2022-03-09
Node chart Hashicorp Vault keys injection

PierreBesson:
Based on https://github.com/paritytech/helm-charts/pull/55 Thanks a lot @dblane-digicatapult for the initial implementation !

Compared to  #55 add the following:
- Add more options for the vault injector to support additional Vault authentication schemes
- Inject node keys and seeds directly into tmp files under `/vault/secrets/`
- Add useful `extraDerivation` option.


charts/node/README.md
```diff
@@ -22,6 +22,31 @@ helm install kusama-node parity/node --set node.chainDataSnapshotUrl=https://ksm
...
+```
+node:
+  vault:
+    keys:
+      - name: aura
+        type: aura
+        scheme: sr25519
+        vaultPath: kv/secret/polkadot-node # path at which the secret is located in Vault
+        vaultKey: aura # key under which the secret value is stored in Vault
+        extraDerivation: "//${HOSTNAME}//aura" # allows to have unique derived keys for each pod of the statefulset
```
dblane-digicatapult, line 43: Very useful.  Definitely an improvement being able to do this vs how we use `node.keys`

aardbol (approved):

BulatSaif (approved):
Good job! 
I think we can mark old `inject-keys` as deprecated, since it is insecure to store key in `values.yaml`

dblane-digicatapult:
> Good job! I think we can mark old `inject-keys` as deprecated, since it is insecure to store key in `values.yaml`

Possibly, there are ways to make it secure e.g. using a GitOps provider like Flux2 to create a configmapRef/secretRef for inputting the values.  So maybe note it but please don't remove it entirely.

PR #62, opened 2022-03-18
[node] fix serviceaccount

BulatSaif:
This PR fixes issue when you have multiple releases with the same serviceAccount, and can not create  `RoleBinding`
```
Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: Role "versi-parachains-service-reader" in namespace "versi" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "versi-tick-extra-501-a": current value is "versi-moonbase-bootnode"
```

### DIFF
--- a/charts/node/templates/statefulset.yaml
+++ b/charts/node/templates/statefulset.yaml
@@ -256,7 +256,7 @@ spec:
               --scheme {{ .scheme }} \
               {{- if .extraDerivation }}
               --suri "$(cat /vault/secrets/{{ .name }}){{ .extraDerivation }}" \
-              {{ else }}
+              {{- else }}
               --suri "/vault/secrets/{{ .name }}" \
               {{- end }}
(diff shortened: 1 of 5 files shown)


charts/node/templates/statefulset.yaml
```diff
@@ -256,7 +256,7 @@ spec:
               --scheme {{ .scheme }} \
               {{- if .extraDerivation }}
               --suri "$(cat /vault/secrets/{{ .name }}){{ .extraDerivation }}" \
-              {{ else }}
+              {{- else }}
```
BulatSaif, line 259: this will remove empty line before `&&` if `extraDerivation` set.

dblane-digicatapult (changes requested):
Add entry for the value to the charts `README.md` ?

PierreBesson (approved):

Issue #29, opened 2021-12-17
Allow fixing the p2p port


PierreBesson:
Automatic node port discovery was introduced in #28, however in some cases the operator will not want / be able to open a large range of ports (30000-32767) on their Kubernetes nodes.

An option should be added to fix the attributed nodePort, however in this case it might be impossible to support more than 1 replica for the statefulset as fixing the port will result in a port conflict for the second replica.

dblane-digicatapult:
When using type LoadBalancer our implementation would vary depending on the cloudProvider being used when we query a service to get the external address which is either an FQDN or an IP,  which we then need to advertise when using `--public-addr`

AWS example:
```
status:
  loadBalancer:
    ingress:
    - hostname: some-loadbalancer-hostname.eu-west-1.elb.amazonaws.com
```

Where as GCP will return:
```
status:
    loadBalancer:
      ingress:
      - ip: 111.111.111.111
```

This complicates how we then advertise, do we want to detect the address's format and then setup our multiaddr format based on if its an IP or FQDN?  The current kubectl image we are using for the init container that queries these services has no `dig` or `hosts` command so reducing an FQDN to either a single or an array of IPs is challenging.

Also currently our naming is very specific to NodePort `node.perNodeServices.createP2pNodePortService` Do we want to change something more generically named for the p2p service? E.G.
```
  perNodeServices:
    createApiService: true
    createP2pService: true
    p2pServiceType: LoadBalancer  # Must be type ClusterIP, NodePort or LoadBalancer
    setPublicAddressToExternal:
      enabled: true
      ipRetrievalServiceUrl: https://ifconfig.io/ip
```

dblane-digicatapult:
@PierreBesson Any comments on above suggestions?

PierreBesson:
For me the renaming you propose with`createP2pService` and `p2pServiceType` make sense. As for supporting FQDN instead of IPs for `--public-addr`, I didn't even know this was possible but I would be fine to add this feature if it is strictly additive to regular IP-based public address.

PierreBesson:
Fixing the p2p ports is now possible with chart `v2.0` using the `node.perNodeServices.relayP2pService.port` and `node.perNodeServices.paraP2pService.port` properties. However fixing the public IP with the Service `loadBalancerIP` still needs to be implemented.

PR #63, opened 2022-03-18
Refactor perNodeServices options and alllow setting the p2p service type

PierreBesson:


charts/node/templates/service.yaml
```diff
@@ -56,8 +60,16 @@ metadata:
...
+  {{- if eq $.Values.node.perNodeServices.p2pServiceType "NodePort" }}
   type: NodePort
   externalTrafficPolicy: Local
+  {{- end }}
+  {{- if eq $.Values.node.perNodeServices.p2pServiceType "ClusterIP" }}
+  type: ClusterIP
+  {{- end }}
+  {{- if eq $.Values.node.perNodeServices.p2pServiceType "LoadBalancer" }}
+  type: LoadBalancer
+  {{- end }}
```
dblane-digicatapult, line 72: Could use `else if eq` for these and skip the unnecessary `-ends`

charts/node/templates/service.yaml
```diff
@@ -77,8 +89,16 @@ metadata:
...
```
dblane-digicatapult, line 101: Could use `else if eq` for these and skip the unnecessary `-ends`

dblane-digicatapult (changes requested):

dblane-digicatapult:
How are you planning to deal with `--public-addr` service address/IP lookup?

I have a rudimentary implementation for my use case (AWS loadbalancer) however this then means advertising `dns4` multiaddr format instead of `ipv4`

The case for loadbalancers is fairly difficult and extremely provider specific and I'm fairly certain there is no "one size fits all" solution.  From memory most of the PaaS providers I tested with advertise the address on either `status.loadbalancer.ingress.ip[]` or `status.loadbalancer.ingress.hostname` So we could try looking up both?

PierreBesson:
> How are you planning to deal with --public-addr service address/IP lookup?

Yes adding the ability to set the loadBalancer IP or hostname would be good. If you want to give it a shot in a following PR it would be very appreciated. We can support any cloud specific way to do this as long as there are users which need this support.

dblane-digicatapult:
> Yes adding the ability to set the loadBalancer IP or hostname would be good. If you want to give it a shot in a following PR it would be very appreciated. We can support any cloud specific way to do this as long as there are users which need this support.

Currently we pass the NodePort `RELAY_CHAIN_P2P_PORT`, `PARA_CHAIN_P2P_PORT` and `EXTERNAL_IP` to the container via a file on the data volume.
Do you want to continue with this approach or should we instead create a configmap which we then do a `kubectl patch` against in the init container? The advantage of using the configmap is that it can be queried by other containers and applications and has potential use for declaring bootnodes and other nodes dynamically.  The disadvantage is that anyone using GitOps tooling to maintain state will have to not overwrite these patches.

Thoughts?

PierreBesson:
>Do you want to continue with this approach or should we instead create a configmap which we then do a kubectl patch against in the init container? The advantage of using the configmap is that it can be queried by other containers and applications and has potential use for declaring bootnodes and other nodes dynamically. The disadvantage is that anyone using GitOps tooling to maintain state will have to not overwrite these patches.

I don't like the approach of using `kubectl patch` directly from the pod. The Kubernetes resource should be the declarative source of truth and should not be modified by init containers. Also GitOps can be pretty much expected. For bootnodes, people should use something like a fixed `nodePort` even though in this case you can only have one replica.

@dblane-digicatapult, it would be better to have this discussion in the issue: https://github.com/paritytech/helm-charts/issues/29

BulatSaif (approved):

PR #80, opened 2022-05-25
Avoid using double the amount of space when downloading a blockstore

bakhtin:

Previous implementation of init container for downloading a blockstore had been using double the amount of disk space. It first downloaded a compressed archive and then unpacked it to the target directory.
This PR fixes it by merging these two steps into one. We stream the compressed byte array produced by `wget` and pipe it further for extraction.

Other quality of life improvements:
- existing implementation for unpacking 7z archives was primarily in place to download snapshots from Polkashots.io. As they migrated from 7z to lz4 compression it no longer worked. I replaced `crazymax/7z` image with [`paritytech/lz4`](https://github.com/paritytech/scripts/pull/425) where `lz4` package is installed to support the new format.
- use `set -eu -pipefail` shell options as a safety net in case of errors in scripts. `-e` will exit immediately on error instead of continuing execution. `-u` will exit on undefined variables. `pipefail` will exit on errors in between the piped programs

### DIFF
--- a/charts/node/templates/statefulset.yaml
+++ b/charts/node/templates/statefulset.yaml
@@ -72,23 +72,19 @@ spec:
           args:
             - -c
             - |
+              set -eu -o pipefail
+              apk add --no-cache curl lz4
               if [ -d "/data/chains/${CHAIN_PATH}/db" ]; then
                 echo "Database directory already exists, skipping chain snapshot download"
               else
                 echo "Downloading chain snapshot"
                 SNAPSHOT_URL="{{ .Values.node.chainDataSnapshotUrl }}"
-                wget -O /data/snapshot ${SNAPSHOT_URL}
-                if [ ! -f /data/snapshot ]; then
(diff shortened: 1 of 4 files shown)


charts/node/templates/statefulset.yaml
```diff
@@ -106,23 +104,19 @@ spec:
           args:
             - -c
             - |
+              set -eu -o pipefail
+              apk add --no-cache wget lz4
```
lazam, line 108: Instead of downloading the wget and lz4, I think it's better to have a generic or dedicated docker image for the init container to have the packages pre-installed. Though the packages are small, downloading the same package when deploying thousand of nodes isn't practical. Also, we might add some packages in the future so it's better to have a docker image for it.

bakhtin: Replaced alpine with paritytech/lz4

BulatSaif (approved):

PR #87, opened 2022-06-14
Parachain Metrics

lazam:
* Expose Prometheus for Parachain node
* Include servicemonitor for parachain metrics
* Updated default values.yml for readability


charts/node/values.yaml
```diff
@@ -54,58 +89,104 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
-# Add extra labels on pods and services
-#extraLabels:
-#  type: rpc
 
+## Deploy a Polkadot node
+## ref: https://wiki.polkadot.network/docs/maintain-sync
```
dblane-digicatapult, line 94: Should probably refer to substrate rather than polkadot given that is the generic upstream product?

charts/node/values.yaml
```diff
@@ -54,58 +89,104 @@ ingress:
...
 node:
   chain: "polkadot"
   command: "polkadot"
   dataVolumeSize: 100Gi
   replicas: 1
   role: full
-  #customChainspecUrl:
-  ## Node may require custom name for chainspec file. Example: moonbeam https://github.com/PureStake/moonbeam/issues/1104#issuecomment-996787548
+  
+  ## Node StorageSotra
```
dblane-digicatapult, line 102: Typo?

charts/node/templates/serviceMonitor.yaml
```diff
@@ -42,3 +42,48 @@ spec:
...
   {{- end}}
 {{- end }}
+---
+{{- if and .Values.node.serviceMonitor.enabled .Values.node.collator.isParachain }}
+{{ $fullname :=  include "node.fullname" . }}
+{{ $serviceLabels :=  include "node.serviceLabels" .  }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ $fullname }}-parachain
```
PierreBesson, line 52: The service monitor shouldn't be labelled `*-parachain`, instead it's the relay-chain (the optionally monitored one) that should be labelled `*-relaychain`. If you are running a parachain, you care less about monitoring the relay-chain but more about monitoring your own chain.

charts/node/templates/statefulset.yaml
```diff
@@ -366,11 +370,15 @@ spec:
                 {{- end }}
                 {{- if .Values.node.tracing.enabled }}
                 --jaeger-agent=127.0.0.1:{{ .Values.jaegerAgent.ports.compactPort }} \
-                {{- end }}
+                {{- end }}
```
PierreBesson, line 376: unnecessary trailing whitespace.

charts/node/values.yaml
```diff
@@ -54,58 +89,102 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
-# Add extra labels on pods and services
-#extraLabels:
-#  type: rpc
 
+## Deploy a substrate node
+## ref: https://docs.substrate.io/v3/getting-started/overview/
```
PierreBesson, line 94: It's strange because this link doesn't provide any help to setup your node. Maybe we can remove it.

charts/node/values.yaml
```diff
@@ -54,58 +89,102 @@ ingress:
...
   customChainspecPath: "/data/chainspec.json"
+  # customChainspecUrl:
   forceDownloadChainspec: false
-  #chainDataSnapshotUrl: "https://dot-rocksdb.polkashots.io/snapshot"
-  #chainDataSnapshotFormat: lz4
-  #chainPath: ""
-  #chainDataKubernetesVolumeSnapshot: ""
-  #chainDataGcsBucketUrl: ""
+
+  ## Configuration to download collator snapshot from remote URL
```
PierreBesson, line 114: Can you change `collator` -> `chain`, as this part works for any kind of node.

charts/node/values.yaml
```diff
@@ -54,58 +89,102 @@ ingress:
...
   vault:
+    ## Vault annotations to use
+    ##
     #authPath: ""
     #authRole: ""
     #authType: ""
     #authConfigType: ""
     #authConfigServiceAccount: ""
+
+    ## Keys to use by the node via vault
```
PierreBesson, line 190: `Keys to use by the node via vault`, this phrasing is hard to get right but it could be something like: `Keys to fetch from Vault and set on the node`. But it's tangential to the current PR.

PierreBesson (changes requested):

PierreBesson:
It looks good, let's do a few more tests and merge.

PR #88, opened 2022-06-14
Add Helm chart testing Workflow

bakhtin:
Added a Gihub Workflow for linting Helm charts and running integration tests by deploying it to a real K8s cluster.

A CI pipeline triggers on every commit to a PR. It only lints and deploys charts that changed. I'm using a [chart-testing](https://github.com/helm/chart-testing) tool from the official Helm repo. It lints a chart according to the rules defined in a [config](chart-testing/lintconf.yaml).
Changed charts are deployed to a k8s cluster created with [`kind`](https://kind.sigs.k8s.io). A chart-testing considers a deployment successful when all the Pods become ready.


.github/workflows/pr.yaml
```diff
@@ -0,0 +1,64 @@
...
+      if: steps.list-changed.outputs.changed == 'true'
+
+    - name: Create kind cluster
+      uses: helm/kind-action@v1.2.0
+      with:
+        node_image: kindest/node:v1.21.10
+      # if: steps.list-changed.outputs.changed == 'true'
+
+    - name: Run chart-testing (install)
+      run: ct install --charts charts/polkadot-introspector --charts charts/substrate-faucet --charts charts/substrate-telemetry
```
PierreBesson, line 62: Would it be possible to automatically include all charts in the `charts/` directory ?

charts/node/values.yaml
```diff
@@ -77,7 +77,7 @@ node:
   collator:
     isParachain: false
     relayChain: polkadot
-  #database: rocksdb
+  database: rocksdb
```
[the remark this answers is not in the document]
bakhtin: A linter throws a warning because you're trying to use an undefined variable in `_helpers.tpl`

```
[ERROR] templates/: template: node/templates/statefulset.yaml:5:20: executing "node/templates/statefulset.yaml" at <include "node.databasePath" .>: error calling include: template: node/templates/_helpers.tpl:91:7: executing "node.databasePath" at <eq .Values.node.database "paritydb">: error calling eq: incompatible types for comparison
```

.github/workflows/pr.yaml
dblane-digicatapult: Last time I looked there was an `--all` flag although the behaviour is dependent on changed versions unlike `--charts` which will run against any chart specified, irrespective of changes or not.

bakhtin: Yep, I'm just testing all charts individually to catch the errors early

PierreBesson (approved):

PR #107, opened 2022-07-01
[Node] add `-x` to initContainers

BulatSaif:
1. We have `echo` to print debug messages in initContainers but sometimes it is not enough to understand which command failed, adding `-x` will fix it.

 
2. Removed `apk add --no-cache wget` since alpine container already has wget

### DIFF
--- a/charts/node/Chart.yaml
+++ b/charts/node/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
-version: 2.6.1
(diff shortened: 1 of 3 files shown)


dblane-digicatapult (changes requested):
Maybe have this optionally set using a helm value like 'debugInitContainers' instead of setting as a default.

BulatSaif:
> Maybe have this optionally set using a helm value like 'debugInitContainers' instead of setting as a default.

Thanks, this is good Idea

ArshamTeymouri (approved):

PierreBesson (approved):

PR #110, opened 2022-07-12
Fix missing Prometheus port

bakhtin:

A Service was missing a Prometheus port. It led to a ServiceMonitor referencing an undefined named port `prometheus` and, consequently, targets not being scraped.

### DIFF
--- a/charts/node/templates/service.yaml
+++ b/charts/node/templates/service.yaml
@@ -25,7 +25,7 @@ spec:
       name: http-rpc
     - port: 9944
       name: websocket-rpc
-    {{- if $.Values.node.serviceMonitor.enabled }}
+    {{- if and $.Values.node.serviceMonitor.enabled (not $.Values.node.perNodeServices.apiService.enabled) }}
     - port: {{ .Values.node.prometheus.port | int }}
       name: prometheus
     {{- end }}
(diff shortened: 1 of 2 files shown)


lazam:
The Prometheus port are exposed per node [here](https://github.com/paritytech/helm-charts/blob/main/charts/node/templates/service.yaml#L66) since we don't want to round-robin the metrics across nodes

bakhtin:
@ArshamTeymouri suggested that instead of PodMonitor we can use headless Service. This way Prometheus can distinguish between different Pods behind the Service and correctly scrape the metrics. Turned out we're already using headless Service in the chart. So my previous changes were almost correct. I added another commit to cover the case when perNodeServices are not enabled.

I tested the metrics collection and it works as expected
image

ArshamTeymouri (approved):

charts/node/templates/service.yaml
```diff
@@ -29,6 +29,10 @@ spec:
     - port: {{ .Values.node.prometheus.port | int }}
       name: prometheus
     {{- end }}
+    {{- if $.Values.node.collator.relayChainPrometheus.enabled }}
+    - port: {{ $.Values.node.perNodeServices.apiService.relayChainPrometheusPort}}
```
dblane-digicatapult, line 33: add `| int` ?

bakhtin: Added it for all other ports in the file

PR #115, opened 2022-07-22
Add chain backup to GCS bucket using an init container

PierreBesson:


BulatSaif:
Thank you for PR, this is very required functionality, but using the init container for that seems odd, is there another place where we can put this logic?  Maybe sidecar container or cronJob?

PierreBesson:
> using the init container for that seems odd, is there another place where we can put this logic? Maybe sidecar container or cronJob?

Sadly I haven't found a better place because it is important that the node be stopped when creating the backup (so we can't use a sidecar container). I agree that it is not ideal that this will create a backup on every restart of the node even if not needed. 
Another possible approach would be to delete the node pod and create a different "node-backup" pod that would attach to the node PVC that we want to backup. This could be developed as an external chart. If you are interested, you can implement this solution.

> is there any way we could make this a more generic backup rsync function with additional flags and options to make it gcs specific?

I agree that it would be much better to have a generic rsync backup feature supporting all kinds of storage locations (S3, rsync, etc.). It would be awesome if you or someone else can contribute this.

BulatSaif (approved):

PR #118, opened 2022-08-02
Add Testnet manager chart

PierreBesson:


charts/testnet-manager/templates/rbac.yaml
```diff
@@ -0,0 +1,76 @@
...
+kind: Role
+metadata:
+  name: {{ $fullName }}-configmap-reader
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
bakhtin, line 22: Should we make it ClusterRole instead so that the manager doesn't require to be deployed in the same namespace as the nodes?

charts/testnet-manager/templates/secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
```
bakhtin, line 1: I'd suggest to also include the option to reference the existing secret as it might be insecure to pass the values directly in the chart.

bakhtin (commented):
Does the chart support running with the default values as is? Otherwise, please provide custom values so that CI can pick it up and test

charts/testnet-manager/templates/rbac.yaml
PierreBesson: I used a ClusterRole in the past but it was a real pain to deploy with helm. In the process of open sourcing it and make it easy to use by third parties, I simplified the setup and made it so that the manager has to be deployed in the same namespace as the node it manages. This way we avoid needing to create additional RBAC resources not managed by the chart, however the manager will only be able to manage nodes in the same namespace which I think is an acceptable constraint for testnets.

PierreBesson:
> Does the chart support running with the default values as is? Otherwise, please provide custom values so that CI can pick it up and test

Yes it can be deployed, although it is not functional as it is missing important config.

bakhtin:
> Yes it can be deployed, although it is not functional as it is missing important config.

Should be fine as long as readiness probes are passing and all the Pods transition to Ready

charts/testnet-manager/templates/secret.yaml
PierreBesson: The way this is is currently set up, it would not be easy to add support for an existing secret. As I would need to implement support for loading an external secret file inthe application code. So I think would be better to tackle this in a future chart upgrade.

bakhtin (approved):

PR #121, opened 2022-08-03
Adds the ability to create custom HPA metrics

dblane-digicatapult:
Templated this out using the following test values

```
...
enabled: true
additionalMetrics:
- type: Pods
  pods:
    metric:
      name: packets-per-second
    target:
      type: AverageValue
      averageValue: 1k
```


bakhtin:
LGTM, thanks. But it needs another version bump. `2.11.2` is already in `main`

dblane-digicatapult:
> LGTM, thanks. But it needs another version bump. `2.11.2` is already in `main`

Done.

bakhtin (approved):

PR #122, opened 2022-08-03
Update HPA apiVersion to v2beta2

bakhtin:

Update HPA apiVersion to v2beta2 as v2 is not yet supported in 1.22.

I initially attempted to implement some smart `apiVersion` resolution as in https://github.com/bitnami/charts/blob/master/template/CHART_NAME/templates/hpa.yaml but it seems that Bitnami's implementation contains a bug. Specifically, if `apiVersion` is set to `v2beta2` an HPA object [definition with `target`](https://github.com/bitnami/charts/blob/master/template/CHART_NAME/templates/hpa.yaml#L29) must used. In contrast, `v2beta1` only supports legacy object definition using [`targetAverageUtilization`](https://github.com/bitnami/charts/blob/master/template/CHART_NAME/templates/hpa.yaml#L27) which is not handled in Bitnami's template.

### DIFF
--- a/charts/node/Chart.yaml
+++ b/charts/node/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
-version: 2.11.1
+version: 2.11.2
 maintainers:
   - name: Parity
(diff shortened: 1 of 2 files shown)


PierreBesson:
Agree to use `v2beta2` as it should continue to work until v1.26.

PR #130, opened 2022-08-17
ExtraArgs capability 

ironoa:
Sometimes extra arguments are needed to be set, such as in [this case](https://github.com/paritytech/substrate-telemetry/pull/491)

The proposed solution might a be a way to do it in a dynamical fashion, not hardcoded

### DIFF
--- a/charts/substrate-telemetry/values.yaml
+++ b/charts/substrate-telemetry/values.yaml
@@ -35,8 +35,7 @@ envVars:
     
 extraArgs:
   shard: {}
-  core:
-    - --expose-node-ips
+  core: {}
   frontend: {}      
 
 serviceMonitor:
--- a/charts/substrate-telemetry/values.yaml
+++ b/charts/substrate-telemetry/values.yaml
@@ -32,10 +32,10 @@ envVars:
     # The frontend docker container makes this available to the UI,
     # so that it knows where to look for feed information:
     SUBSTRATE_TELEMETRY_URL: "wss://core-service-domain.com"
-    
+
 extraArgs:
   shard: {}
-  core: {}    
+  core: {}
 
 serviceMonitor:
   # Only core service has Prometheus metrics exposed at the moment.
--- a/charts/substrate-telemetry/values.yaml
+++ b/charts/substrate-telemetry/values.yaml
@@ -32,6 +32,12 @@ envVars:
     # The frontend docker container makes this available to the UI,
     # so that it knows where to look for feed information:
     SUBSTRATE_TELEMETRY_URL: "wss://core-service-domain.com"
+    
+extraArgs:
+  shard: {}
+  core:
+    - --expose-node-ips
+  frontend: {}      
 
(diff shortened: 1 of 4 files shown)


charts/substrate-telemetry/values.yaml
```diff
@@ -32,6 +32,12 @@ envVars:
     # The frontend docker container makes this available to the UI,
     # so that it knows where to look for feed information:
     SUBSTRATE_TELEMETRY_URL: "wss://core-service-domain.com"
+    
+extraArgs:
+  shard: {}
+  core:
+    - --expose-node-ips
```
PierreBesson, line 39: I think we should not set `--expose-node-ips` as default here. We should respect defaults from the upstream project. Moreover this might be incompatible with older versions of the docker image.

@lexnv Was it decided to enable this as default ?

charts/substrate-telemetry/values.yaml
```diff
@@ -32,6 +32,12 @@ envVars:
     # The frontend docker container makes this available to the UI,
     # so that it knows where to look for feed information:
     SUBSTRATE_TELEMETRY_URL: "wss://core-service-domain.com"
+    
+extraArgs:
+  shard: {}
```
PierreBesson, line 37: The variables `extraArgs.shard` and `extraArgs.frontend` have been defined but are not used in the template. I think you might have just forgotten to `git add` them.

PierreBesson (changes requested):

charts/substrate-telemetry/values.yaml
jsdw: I agree that we shouldn't have that enabled by default. Let's only expose node IPs when we need to.

lexnv: As James mentioned we should not enable this by default. It only needs enablement for W3F 1kV program

ironoa: I completely agree with not embedding `--expose-node-ips` as default here, I just wanted to show you the use case that brought me to make the modification.

charts/substrate-telemetry/values.yaml
ironoa: I added the usage in the shard template. Instead, I removed `extraArgs.frontend` since the frontend app does not expect any argument to be passed (it just uses environment variables and therefore rely on `envVars.fronted`).
wdyt @PierreBesson ?

PierreBesson: Perfect

PierreBesson:
@ironoa Can you bump the chart version ? This is required for auto-publishing.

PierreBesson (approved):

ironoa:
> @ironoa Can you bump the chart version ? This is required for auto-publishing.

Thanks for having taken care of that, minor bump makes sense to me

PierreBesson:
Thanks for your contribution @ironoa !

PR #143, opened 2022-08-26
Add option to mount the node keystores volumes in memory

### LINKED ISSUE #60: Allow mounting a Kubernetes emptyDir at the keystore path
When inserting keys into a substrate node those will end up in the `/data/.../chains/chain_name/keystore` folder which in our setup is stored in the "data" Kubernetes volume. In a secure setup, we don't want this data volume to contain our private keys so we should mount a tmpfs (ie [Kubernetes volume emptyDir type](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir)) on this path.
This will prevent keys from being persisted in the data disk after having been sourced from Hashicorp Vault or other secure places.

PierreBesson:
Fix #60

Needs further testing


bakhtin:
@PierreBesson I updated your PR with some small changes and added a `sizeLimit` parameter for the in-memory volumes. I successfully tested the chart locally so we're good to merge it

charts/node/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
-version: 3.2.3
+version: 3.3.3
```
lazam, line 5:
```suggestion
version: 3.3.0
```

BulatSaif (approved):

lazam (approved):

charts/node/templates/statefulset.yaml
```diff
@@ -640,6 +640,22 @@ spec:
             secretName: {{ $fullname }}-{{ .type }}
             defaultMode: 0400
       {{- end }}
+      {{- if .Values.node.chainKeystore.mountInMemory.enabled }}
+        - name: chain-keystore
+          emptyDir:
+            medium: "Memory"
+            {{- if .Values.node.chainKeystore.mountInMemory.sizeLimit }}
+            sizeLimit: {{ .Values.node.chainKeystore.mountInMemory.sizeLimit }}
+            {{- end }}
```
dblane-digicatapult, line 649:
```suggestion
            {{- if and (.Values.node.chainKeystore.mountInMemory.sizeLimit) (semverCompare ">= 1.22-0") }}
            sizeLimit: {{ .Values.node.chainKeystore.mountInMemory.sizeLimit }}
            {{- end }}
```

This uses the capabilities functionality in helm to determine if the minimum version of kubernetes is sufficient to support `sizeLimit`

dblane-digicatapult: I would advise testing this as I am just doing it from memory - untested.

bakhtin: Added a version check when sizeLimit is set

PierreBesson:
@bakhtin your changes LGTM

PR #160, opened 2022-09-14
[Substrate-faucet] Existing configmap and secret support

PierreBesson:


charts/substrate-faucet/values.yaml
```diff
@@ -27,11 +27,11 @@ bot:
     repository: paritytech/faucet-bot
     tag: latest
     pullPolicy: Always
-
+  existingConfigMap: ''
+  existingSecret: ''
```
dblane-digicatapult, line 31:
```suggestion
  existingConfigMap: ""
  existingSecret: ""
```
use double quotes for consistency?

aardbol (approved):

charts/substrate-faucet/values.yaml
```diff
@@ -12,10 +12,10 @@ server:
     repository: paritytech/faucet-server
     tag: latest
     pullPolicy: Always
-
+  existingConfigMap: ''
```
aardbol, line 15: You don't want double quotes here too for consistency?

Issue #142, opened 2022-08-26
Revamp node flags settings using charts values


PierreBesson:
When using the node helm chart, it is really hard to figure out whether flags should be set in the chart values or inside `node.flags`. The situation can be improved:
- [x]  List common substrate flags option and allow configuring them with chart values (--unsafe-rpc-external, --telemetry-url, --log). We shouldn't include more esoteric flags (polkadot/cumulus specific)
- [x] The `pruning` and `database` options should be under `chainData` and `relayChainData`.
- [x] Preferentially use default values (ie. not setting the flags when value=null)
- [x] relayChainFlags should have `--name=${POD_NAME}"` by default and the same telemetry URLs as the main chain.
- [x] Add script on node startup to check if any passed flags is in the list of flags that are handled by chart values and fail if it's the case
- [x] Also forbid to set flags which don't make sense for the chart: `--ws-port` and `--rpc-port`

PierreBesson:
Should we completely revamp the collator options ? It would be a pretty drastic change but it will make things more readable.

We could change this:
```
  collator:
    isParachain: true
    relayChain: rococo
    relayChainDatabase: paritydb
    relayChainPruning: 1000
    relayChainCustomChainspecUrl: "https://paritytech.github.io/chainspecs/versi/relaychain/chainspec.json"
    relayChainData:
      chainPath: rococo_v2_2
      GCSBucketUrl: gs://rococo-blockstore-backups/rococo-paritydb-prune
      storageClass: ssd-csi
      volumeSize: 30G
    relayChainKeystore:
      mountInMemory: true
```
Into:
```
isParachain: true
collatorRelayChain:
  chain: rococo
  database: paritydb
  pruning: 1000
  customChainspecUrl:
  chainData:
      chainPath: rococo_v2_2
      GCSBucketUrl: gs://**
      storageClass: ssd-csi
      volumeSize: 50G
  chainKeystore:
      mountInMemory: true
```

This has the the big advantage of keeping same options names to configure the collator relay chain just indented under `collatorRelayChain`

gabreal:
looks good to me. i'd suggest to add those flags as variables that have consequences in the configuration of the deployment. for ones that don't affect the deployment code e.g. `sync=debug` they can be easily added via flags.

BulatSaif:
I would like to add: 
[...]

PR #161, opened 2022-09-15
Node chart v4 : values revamp

PierreBesson:
This is a big node chart values revamp which introduces a lot of breaking changes.

Close: https://github.com/paritytech/helm-charts/issues/142


charts/node/templates/_helpers.tpl
```diff
@@ -88,9 +88,16 @@ Create the name of the service account to use
...
 {{- else }}
 {{- "db" }}
 {{- end }}
 {{- end }}
+
+{{/*
+Define a regex matcher to check if the passed node flags passed are already managed by the chart
+*/}}
+{{- define "node.chartManagedFlagsRegex" -}}
+{{- "(\\W|^)(--name|--base-path|--chain|--validator|--collator|--light|--database|--pruning|--prometheus-external|--prometheus-port|--node-key|--wasm-runtime-overrides|--jaeger-agent|--rpc-external|--unsafe-rpc-external|--ws-external|--unsafe-ws-external|--rpc-methods|--rpc-cors|--rpc-port|--ws-port)(\\W|$)" }}
```
dblane-digicatapult, line 102: Maybe move this to _helpers.tpl ?

charts/node/templates/statefulset.yaml
```diff
@@ -394,7 +417,7 @@ spec:
           args:
             - -c
             - |
-              set -eu
+              set -eux
```
dblane-digicatapult, line 423: Set the x conditionally as we do in init containers?

charts/node/templates/_helpers.tpl
BulatSaif: It is already in `charts/node/templates/_helpers.tpl`

charts/node/templates/statefulset.yaml
BulatSaif: Fixed, added option to disable/enable this behavior

charts/node/templates/statefulset.yaml
```diff
@@ -288,6 +289,28 @@ spec:
           securityContext:
             runAsUser: 0
         {{- end }}
+        {{- if .Values.node.persistGeneratedNodeKey }}
+        - name: persist-generated-node-key
```
BulatSaif, line 293: how it was generated before?

charts/node/templates/statefulset.yaml
```diff
@@ -121,9 +122,9 @@ spec:
               fi
           env:
             - name: RELAY_SNAPSHOT_FORMAT
-              value: {{ default "tar" .Values.node.collator.relayChainData.snapshotFormat }}
+              value: {{ default "tar" .Values.node.collatorRelayChain.chainData.snapshotFormat }}
             - name: RELAY_CHAIN_PATH
-              value: {{ default .Values.node.collator.relayChain .Values.node.collator.relayChainData.chainPath }}
+              value: {{ default .Values.node.collatorRelayChain.chain.Values.node.collatorRelayChain.chainData.chainPath }}
```
BulatSaif, line 127:
```suggestion
              value: {{ default .Values.node.collatorRelayChain.chain .Values.node.collatorRelayChain.chainData.chainPath }}
```

charts/node/templates/statefulset.yaml
```diff
@@ -226,7 +227,7 @@ spec:
               fi
           env:
             - name: RELAY_CHAIN_PATH
-              value: {{ default .Values.node.collator.relayChain .Values.node.collator.relayChainData.chainPath }}
+              value: {{ default .Values.node.collatorRelayChain.chain.Values.node.collatorRelayChain.chainData.chainPath }}
```
BulatSaif, line 230:
```suggestion
              value: {{ default .Values.node.collatorRelayChain.chain .Values.node.collatorRelayChain.chainData.chainPath }}
```

charts/node/README.md
```diff
@@ -69,6 +69,48 @@ node:
...
+- `node.collator.relayChainCustomChainspecUrl` -> `node.collatorRelayChain.customChainspecUrl`
+- `node.collator.relayChainFlags` -> `node.collatorRelayChain.flags`
+- `node.collator.relayChainData.*` -> `node.collatorRelayChain.chainData.*`
+- `node.collator.relayChainPruning` -> `node.collatorRelayChain.chainData.pruning`
+- `node.collator.relayChainDatabase` -> `node.collatorRelayChain.chainData.database`
+- `node.collator.relayChainKeystore.*` -> `node.collatorRelayChain.chainKeystore.*`
+- `node.collator.relayChainPrometheus.*` -> `node.collatorRelayChain.prometheus.*`
+
+The following flags are now invalid if they were previously set in `node.flags` or `node.collator.relayChainFlags`.
+An error will be thrown if any of those flags is set directly.
```
bakhtin, line 90:
```suggestion
An error will be thrown if any of those flags are set directly.
```

charts/node/templates/_helpers.tpl
```diff
@@ -88,9 +88,16 @@ Create the name of the service account to use
...
-{{- if eq .Values.node.database "paritydb" }}
+{{- if eq .Values.node.chainData.database "paritydb" }}
 {{- "paritydb" }}
 {{- else }}
 {{- "db" }}
 {{- end }}
 {{- end }}
+
+{{/*
+Define a regex matcher to check if the passed node flags passed are already managed by the chart
```
bakhtin, line 99:
```suggestion
Define a regex matcher to check if the passed node flags are managed by the chart already
```

charts/node/templates/statefulset.yaml
```diff
@@ -288,6 +289,28 @@ spec:
...
+              set -eu {{ if .Values.initContainer.debug }}-x{{ end }}
+              NODE_KEY_PATH="/keystore/node-key"
+              if [ -f "${NODE_KEY_PATH}" ]; then
+                echo "Node key already exists, skipping node key generation"
+              else
+                {{ $.Values.node.command }} key generate-node-key --file ${NODE_KEY_PATH} \
+                && echo "Generate node key into Keystore" \
+                || echo "Failed to insert key into Keystore."
+              fi
+              NODE_PEER_ID="$(cat ${NODE_KEY_PATH} | {{ .Values.node.command }} key inspect-node-key)"
```
bakhtin, line 308: This may expose the node key in the logs. Make sure to use `key inspect-node key --file <node_key_path>`

charts/node/templates/statefulset.yaml
bakhtin: I guess the node key is generated when the node starts. So we did not need to explicitly generate it. Not sure how `persistGeneratedNodeKey` worked though

charts/node/templates/statefulset.yaml
```diff
@@ -453,8 +476,21 @@ spec:
                 --prometheus-external \
                 --prometheus-port {{ .Values.node.prometheus.port }} \
                 {{- end }}
-                {{- if .Values.node.collator.isParachain }}
-                {{- if (or $.Values.node.perNodeServices.paraP2pService.enabled) }}
+                {{- /*
+                Those unsafe flags are required to expose the RPC interfaces.
+                This a not security risk if those are not exposed publicly.
```
bakhtin, line 481:
```suggestion
                It is not a security risk unless they are exposed publicly.
```

charts/node/templates/statefulset.yaml
```diff
@@ -453,8 +476,21 @@ spec:
                 --prometheus-external \
                 --prometheus-port {{ .Values.node.prometheus.port }} \
                 {{- end }}
-                {{- if .Values.node.collator.isParachain }}
-                {{- if (or $.Values.node.perNodeServices.paraP2pService.enabled) }}
+                {{- /*
+                Those unsafe flags are required to expose the RPC interfaces.
```
bakhtin, line 480:
```suggestion
                The unsafe flags are required to expose RPC interfaces.
```

bakhtin (changes requested):

charts/node/templates/statefulset.yaml
BulatSaif: Not all implimetation of substrate node will have `key` option, we may need to create separate container  with `subkey`
```
error: Found argument 'key' which wasn't expected, or isn't valid in this context

USAGE:
    malus dispute-ancestor [OPTIONS]

For more information try --help
```

PierreBesson: Yes but this is inconvenient as there are no official binary releases or docker image of Subkey.

PierreBesson: `persistGeneratedNodeKey` was broken for a few releases as the code disappeared during a previous refactor.

BulatSaif: we can try to use this docker image, it even had update 1 hour ago: 
https://hub.docker.com/r/parity/subkey/tags

charts/node/templates/statefulset.yaml
PierreBesson: I think this didn't exist before, but you are right is the better option.

BulatSaif (approved):

bakhtin (approved):

PR #162, opened 2022-09-15
Dump session keys to stdout

bakhtin:

Add Python script that dumps session keys to stdout when keys are injected through Vault. The motivation behind this is to obtain a session key for later use in `setKeys()` RPC call.

### DIFF
--- a/charts/node/Chart.yaml
+++ b/charts/node/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
-version: 3.3.0
+version: 3.3.1
 maintainers:
   - name: Parity
     url: https://github.com/paritytech/helm-charts
--- a/charts/node/templates/statefulset.yaml
+++ b/charts/node/templates/statefulset.yaml
@@ -359,6 +359,54 @@ spec:
           volumeMounts:
             - mountPath: /keystore
               name: chain-keystore
+        - name: dump-session-keys
+          image: docker.io/python:3.10.7-slim
+          command: [ "python" ]
+          args:
+            - -c
+            - |
+              import os
+              import logging
+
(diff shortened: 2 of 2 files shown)


BulatSaif:
I like the idea of exposing the session key,  it can be useful (example: expose it as k8s labels and use it in monitoring), but your use case looks very limited and I see the following problems:
1. Node can have several session keys, your script will publish the last one, but it may not be the active key. 
2. It will work only with an injected key (which is not a best practice), If the key is rotated you need to restart the pod.

If you already have keys in vault you can generate session key from private keys and store it in vault. 

I don't think we should include this script in public helm, maybe you can use the approach from staking-miner helm chart? 
```
      initContainers:
        {{- if .Values.initContainers }}
          {{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}
        {{- end }}
```

bakhtin:
I agree that the use-case is narrow and only applicable to certain deployment scenarios. But since we don't have requests to cover more scenarios the current implementation can serve well to support at least some of them.
I can refactor the PR to rely on the optional `initContainers` instead of introducing a new script.

Are you aware of a better way to obtain a session key in runtime?

PierreBesson:
@BulatSaif
> Node can have several session keys, your script will publish the last one, but it may not be the active key.

The only way to get the active key is to query it on-chain but I don't think it matters here as this script is intended for setting up new session keys.

> It will work only with an injected key (which is not a best practice), If the key is rotated you need to restart the pod.

Arguably this will only be needed for injected keys as the `rotate_key` RPC call already outputs the session key in response.

> If you already have keys in vault you can generate session key from private keys and store it in vault.

It's true that session key could be generated directly from the Vault private keys without reading the keystore filesystem but I think this script will be very convenient to have as part of the chart even though I agree that it doesn't appear like the cleanest solution.

bakhtin:
Closing this in favor of https://github.com/paritytech/helm-charts/pull/164. I'll add the equivalent functionality using `iniContainers` defined in the deployment values.

PR #169, opened 2022-10-05
[Node] Replace Kubectl Image in init-container

lazam:
With the recent changes in the bitnami/kubectl, it removes the package `curl` which breaks the retrieval of port. It's more beneficial to use our own image in case we need to add more packages and avoid having unexpected package removal.

### DIFF
--- a/charts/node/values.yaml
+++ b/charts/node/values.yaml
@@ -38,8 +38,8 @@ initContainer:
 ##
 kubectl:
   image:
-    repository: bitnami/kubectl
-    tag: latest
+    repository: paritytech/kubetools
+    tag: kubectl
 
 ## Used to sync chain from/to GCS bucket
 ##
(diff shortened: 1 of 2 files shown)


charts/node/values.yaml
```diff
@@ -38,8 +38,8 @@ initContainer:
 ##
 kubectl:
   image:
-    repository: bitnami/kubectl
-    tag: latest
+    repository: paritytech/kubetools
+    tag: kubectl
```
bakhtin, line 42: Can we have multiple versioned tags? E.g., `paritytech/kubetools:kubectl-1.22`, `paritytech/kubetools:kubectl-latest`

lazam: `kubectl-latest` is `kubectl`, while specific version is `kubectl-$KUBE_VERSION`.

BulatSaif (approved):

PierreBesson:
Sorry I should have merged this before the node v4 breaking changes so it would be available in both v3 and v4.

PierreBesson (approved):

PR #177, opened 2022-10-10
add parachain p2p ws port

ArshamTeymouri:
add support for parachain p2p ws port


charts/node/templates/service.yaml
```diff
@@ -150,6 +150,11 @@ spec:
     - name: p2p
       port: {{ $.Values.node.perNodeServices.paraP2pService.port | int }}
       targetPort: 30334
+    {{- if $.Values.node.perNodeServices.paraP2pService.ws.enabled }}
+    - name: ws
+      port: {{ $.Values.node.perNodeServices.paraP2pService.ws.port | int }}
+      targetPort: 30334
```
aardbol, line 156: Same targetPort as the p2p one, is that right?

ArshamTeymouri: Yes. the p2p port is 30333 whereas p2p-ws is 30334, both for relaychain and parachain services.

charts/node/values.yaml
```diff
@@ -338,6 +338,10 @@ node:
         enabled: false
         hostname: example.com
         ttl: 300
+      # If enabled, additionally expose WebSocket port. Useful for bootnodes
+      ws:
+        enabled: false
+        port: 30334
```
PierreBesson, line 344: This port needs to be different from `paraP2pService.port` (TCP p2p) and `relayP2pService.port`. I would suggest 30444.

charts/node/templates/service.yaml
ArshamTeymouri: sorry I checked again, you're right. The port should be changed.
@PierreBesson the parachain p2p service port is set to 30334, while this port for relaychain is 30333. could we change the port in parachain to 30333 as well as those are two separate services and could have the same port set?
otherwise I would keep 30334 for parachain-p2p and 30335 for parachain-p2p-ws.

charts/node/values.yaml
ArshamTeymouri: I updated the ports:
- relaychain-p2p: 30333 (remained unchanged)
- relaychain-p2p-ws: 30334 (remained unchanged)
- parachain-p2p: 30444
- parachain-p2p-ws: 30445

aardbol: Never mind, it seems that they are fixed before I posted

aardbol: Related: https://github.com/paritytech/helm-charts/issues/178

charts/node/templates/service.yaml
```diff
@@ -114,7 +114,7 @@ spec:
     {{- if $.Values.node.perNodeServices.relayP2pService.ws.enabled }}
     - name: ws
       port: {{ $.Values.node.perNodeServices.relayP2pService.ws.port | int }}
-      targetPort: 30334
+      targetPort: 30444
```
PierreBesson, line 117: I would set it to the opposite, it should be 29444. When we deploy a parachain node, the relay port is "secondary" and honestly few people would care about it.

charts/node/templates/statefulset.yaml
```diff
@@ -490,13 +497,20 @@ spec:
                 --rpc-methods=unsafe \
                 {{- end }}
                 {{- if .Values.node.isParachain }}
+                --listen-addr=/ip4/0.0.0.0/tcp/29333 \
```
PierreBesson, line 500: 30333

PierreBesson (changes requested):

charts/node/templates/service.yaml
ArshamTeymouri: Agree, but for non-parachain nodes, the relaychain would be primary chain. see https://github.com/paritytech/helm-charts/issues/178#issuecomment-1273418120, please. let me know about the final decision.

charts/node/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
-version: 4.0.2
+version: 4.0.3
```
lazam, line 5: please bump the minor version

charts/node/README.md
```diff
@@ -230,14 +230,16 @@ If you're running a collator node:
...
 | `node.perNodeServices.relayP2pService.port`                               | Port to use to expose the relay-chain P2P Service externally                                                                                                                                                                                         | `9944`                                                              |
 | `node.perNodeServices.relayP2pService.ws.enabled`                            | If true, creates a relay-chain P2P Service for every node in the statefulset exposing the WebSocket P2P endpoint                                                                                                                          | `false`                                                              |
-| `node.perNodeServices.relayP2pService.ws.port`                            | Port to use to expose the relay-chain WebSocket P2P Service externally                                                                                                                          | `30334`                                                              |
```
lazam, line 233: Please add the following breaking changes in the README when moving to this version

PierreBesson:
Maybe we can revert all ports change, this was not my intention to push this in this unrelated MR.

charts/node/templates/service.yaml
PierreBesson: final decision is to leave it aside for now and use `30334` for p2p ports other than relay-tcp.

ArshamTeymouri:
I reverted all the port changes to be followed later in a separate PR as per https://github.com/paritytech/helm-charts/issues/178

PierreBesson (approved):

charts/node/Chart.yaml
PierreBesson: Should be minor instead of fix so: 4.1.0

PierreBesson (approved):

PR #182, opened 2022-10-18
Enhance paraP2pServices conditions

ArshamTeymouri:


charts/node/templates/statefulset.yaml
```diff
@@ -449,12 +451,17 @@ spec:
...
+              {{- if eq .Values.node.perNodeServices.paraP2pService.type "NodePort" }}
               PARA_CHAIN_P2P_PORT="$(cat /chain-data/para_chain_p2p_port)"
-              echo "PARA_CHAIN_P2P_PORT=${PARA_CHAIN_P2P_PORT}"
               {{- if .Values.node.perNodeServices.paraP2pService.ws.enabled }}
               PARA_CHAIN_P2P_PORT_WS="$(cat /chain-data/para_chain_p2p_port_ws)"
-              echo "PARA_CHAIN_P2P_PORT_WS=${PARA_CHAIN_P2P_PORT_WS}"
               {{- end }}
+              {{- else }}
+              PARA_CHAIN_P2P_PORT={{ $.Values.node.perNodeServices.paraP2pService.port | quote }}
+              PARA_CHAIN_P2P_PORT_WS={{ $.Values.node.perNodeServices.paraP2pService.ws.port | quote }}
```
PierreBesson, line 461: Shouldn't this be wrapped with ` {{- if .Values.node.perNodeServices.paraP2pService.ws.enabled }}` similar to the block above ?

ArshamTeymouri: good catch! resolved.

PierreBesson (approved):

PR #186, opened 2022-10-27
Chore/upgrade actions

dblane-digicatapult:
This moves several github actions to newer versions.

GHA are deprecating actions running on NodeJS version 12.  This means actions must be updated to using NodeJS 16 where applicable.

Additionally they are deprecating usage of the `::set-output` notation and replacing it with env files (which also necessitates an upgrade on several actions).

The only action that currently doesn't support the env file method for setting outputs currently used in this repository is `Azure/setup-helm` Although this looks to just be a case of upgrading the `@actions/*` deps and looks to be a very quick fix/release when they choose to do so.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -20,21 +20,21 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
     - name: Set up Helm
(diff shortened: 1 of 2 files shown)


PierreBesson (approved):

PierreBesson:
Thank you for your contribution @dblane-digicatapult !

PR #190, opened 2022-11-22
feat(node): add keysExistingSecrets & extraConfigmapMounts

michalziobro:
This is a draft pull request with proposed changes that we found useful during development.
Features:
- possibility to add node keys from already existing k8s secrets. This gives the user more flexibility on k8s secrets management. It's not limited to Vault integration only, secrets could be created manually, using 1Password integration, Bitnami Sealed Secrets or any other preferred.
- mounting ConfigMaps as volumes to node pods. This is a quite generic feature offered by many helm charts. We found it useful for adding config files for custom substrate based nodes.

### DIFF
--- a/charts/node/templates/statefulset.yaml
+++ b/charts/node/templates/statefulset.yaml
@@ -345,6 +345,37 @@ spec:
             - mountPath: /var/run/secrets/{{ .type }}
               name: {{ .type }}
           {{- end }}
+        {{ else if .Values.node.keysExistingSecrets }}
+        - name: inject-existing-keys
+          image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+          command: [ "/bin/sh" ]
+          args:
+            - -c
+            - |
+              set -eu {{ if .Values.initContainer.debug }}-x{{ end }}
+              {{- range $keys := .Values.node.keysExistingSecrets }}
+              if [ ! -f /var/run/secrets/{{ $keys }}/type ]; then
+                 echo "Error: File /var/run/secrets/{{ $keys }}/type does not exist"
+                 exit 1
+              fi
+              {{ $.Values.node.command }} key insert \
+              --keystore-path /keystore \
+              --key-type $(cat /var/run/secrets/{{ $keys }}/type) \
+              --scheme $(cat /var/run/secrets/{{ $keys }}/scheme) \
+              --suri /var/run/secrets/{{ $keys }}/seed \
+              && echo "Inserted key {{ $keys }} into Keystore" \
+              || echo "Failed to insert key {{ $keys }} into Keystore."
+              {{- end }}
+          env:
+            - name: CHAIN
+              value: {{ .Values.node.chain }}
+          volumeMounts:
+            - mountPath: /keystore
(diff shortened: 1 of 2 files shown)


ddorgan:
@michalziobro can you sign the CLA please.

michalziobro:
@ddorgan signed

charts/node/values.yaml
```diff
@@ -227,6 +227,9 @@ node:
   #   seed: "//Alice"
   #   extraDerivation: //babe
 
+  ## Inject keys from already existing kubernetes secrets
+  keysExistingSecrets: []
```
PierreBesson, line 231: Can this be renamed to `node.existingSecrets.keys` to follow the same convention as `node.vault.keys`.
Can you also implement `node.existingSecrets.nodeKey` for loading the node key from a secret. This can be a more secure alternative to `node.customNodeKey` which passes the key as plain text.

michalziobro: Hi @PierreBesson , that's a good idea, added this & changed naming. I also bumped version in Chart.yaml

charts/node/templates/statefulset.yaml
```diff
@@ -655,11 +688,20 @@ spec:
             - mountPath: /relaychain-keystore
               name: relaychain-keystore
             {{- end }}
+          {{- range $keys := .Values.node.extraConfigmapMounts }}
```
PierreBesson, line 691: The $keys variable is unused here, you can replace by;

```suggestion
          {{- range .Values.node.extraConfigmapMounts }}
```

charts/node/templates/statefulset.yaml
```diff
@@ -743,6 +785,12 @@ spec:
       {{- toYaml . | nindent 8 }}
       {{- end }}
       volumes:
+      {{- range $keys := .Values.node.extraConfigmapMounts }}
```
PierreBesson, line 788: The $keys variable is unused here, you can replace by;

```suggestion
      {{- range .Values.node.extraConfigmapMounts }}
```

charts/node/values.yaml
```diff
@@ -227,6 +227,16 @@ node:
   #   seed: "//Alice"
   #   extraDerivation: //babe
 
+  ## Inject keys from already existing kubernetes secrets
+  existingSecrets:
+    ## List of keys to be added to keystore
```
PierreBesson, line 232: Can you document the secret format to be used here.
1. State that the value of `node.existingSecrets.keys` should be a list of Kubernetes secret names.
2. Document that the secret should contains the 3 keys: `type`, `scheme` and `seed`.

PierreBesson (changes requested):

PierreBesson:
If possible could you also add the new options to `charts/node/README.md` ? I can also do it later if you are unsure of how to do this.

michalziobro:
> If possible could you also add the new options to `charts/node/README.md` ? I can also do it later if you are unsure of how to do this.

Hi @PierreBesson , are you using some tool like helm-docs to generate Readme?

PierreBesson (approved):

PR #194, opened 2022-12-07
[parity-bridges-common] rename parity-bridge-common -> bridges-common-relay

BulatSaif:
As been reported by [developers](https://github.com/paritytech/devops/issues/1935#issuecomment-1339045499):
> and one small nit, original repo is parity-bridges-common but here in helm it is used parity-bridge-common,
so if we want to be aligned and perfectionists to avoid futuru confusions, we could rename everything (subdirectory, helm stuff) to parity-bridges-common

the helm chart should have the same name as  repo  https://github.com/paritytech/parity-bridges-common to avoid futuru confusions


BulatSaif:
If we are going to rename I am suggesting the `substrate-relay` - the name of the image.

lazam:
> If we are going to rename I am suggesting the `substrate-relay` - the name of the image.

The reason I used `parity-bridge-common` was because the source code is from this repo https://github.com/paritytech/parity-bridges-common. If we use `substrate-relay` as a name, then it might be confusing that we still refer to this old repo https://github.com/paritytech/substrate-bridge-relay. 

If we're planning to rename to `substrate-relay`, then we should also update the readme and add the details there.

BulatSaif:
> > If we are going to rename I am suggesting the `substrate-relay` - the name of the image.
> 
> The reason I used `parity-bridge-common` was because the source code is from this repo https://github.com/paritytech/parity-bridges-common. If we use `substrate-relay` as a name, then it might be confusing that we still refer to this old repo https://github.com/paritytech/substrate-bridge-relay.
> 
> If we're planning to rename to `substrate-relay`, then we should also update the readme and add the details there.

Based on @lazam  comment I am proposing a new name:
`parity-bridges-common-relay` 
but we also need to rename the docker image to have consistency, I created an issue in the app repo:
https://github.com/paritytech/parity-bridges-common/issues/1708

BulatSaif:
New name helm name is  `bridges-common-relay`

the git repository will be also renamed:
`paritytech/parity-bridges-common  -> paritytech/bridges-common`

bakhtin (approved):

PR #213, opened 2023-02-06
[Faucet] Allow disabling bot and enabling server external access via ingress

PierreBesson:
- Add configuration for faucet server external access introduced in https://github.com/paritytech/substrate-matrix-faucet/pull/192
- Add optional faucet ingress

### DIFF
--- a/charts/substrate-faucet/templates/servicemonitor.yaml
+++ b/charts/substrate-faucet/templates/servicemonitor.yaml
@@ -2,24 +2,12 @@
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  name: {{ .Release.Name }}-bot
-spec:
-  selector:
-    matchLabels:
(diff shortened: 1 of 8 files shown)


charts/substrate-faucet/templates/servicemonitor.yaml
```diff
@@ -1,25 +1,13 @@
 {{- if .Values.serviceMonitor.enabled }}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
-metadata:
-  name: {{ .Release.Name }}-bot
-spec:
-  selector:
-    matchLabels:
-    {{- include "faucet.botSelectorLabels" . | nindent 8 }}
```
BulatSaif, line 9: Is bot don't have metrics?

PierreBesson: Yes, this was actually failing.

PR #219, opened 2023-03-04
*added support for offchain indexing to node helm chart

okalenyk:


PierreBesson:
Thanks for the contribution @okalenyk. As a standard feature of substrate, this should be added to the helm chart.

charts/node/values.yaml
```diff
@@ -449,6 +449,10 @@ node:
     #   mountPath: /etc/config/
     #   readOnly: true
 
+  # Enable Offchain Indexing.
+  # https://docs.substrate.io/fundamentals/offchain-operations/
+  offchainIndexing: false
```
PierreBesson, line 454: Can you please rename this to `enableOffchainIndexing`  to be consistent with the flag.

PierreBesson (changes requested):
Can you also add the `--enable-offchain-indexing` flag to the regex in https://github.com/paritytech/helm-charts/blob/main/charts/node/templates/_helpers.tpl#L103
This will validate that the flag is not already passed via `node.flags`.

charts/node/values.yaml
okalenyk: Done. Thanks for reviewing.

PierreBesson:
Thank you can you also bump the minor version of the chart for the auto-release ?

okalenyk:
> Thank you! Can you also bump the minor version of the chart for the auto-release ?

Sure. Bumped to 4.4.1.

PierreBesson (approved):

PR #235, opened 2023-04-07
Correctly use cluster's default for default storage class

### LINKED ISSUE #201: [node] Don't set storage class to "default" in order to use the real default cluster storage class
In the node helm-chart, we force set the `node.chainData.storageClass" to "default", this prevents Kubernetes from using the cluster defined default storage class if this one is not named `default`.
The solution is to not set the storageClass line in the YAML template when the value is unset.

Reference: https://kubernetes.io/docs/concepts/storage/storage-classes/

PierreBesson:
Fix #201

### DIFF
--- a/charts/node/Chart.yaml
+++ b/charts/node/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
-version: 4.4.1
+version: 4.5.0
 maintainers:
   - name: Parity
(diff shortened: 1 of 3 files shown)


BulatSaif:
I remember that it was common to set the default value to `"-"`, I think it was done to avoid issues with rendering templates or kubernetes had different behavior for `null` and empty string  example: 
https://github.com/helm/charts/pull/2005/files

I search the internet but did not find any best practice or explanation for why `"-"`was used like that.

If your solution works we can merge it.

BulatSaif (approved):

PR #236, opened 2023-04-17
Using custom mounted folders for node keys instead of tmp

okalenyk:
Problem details: [cannot bind unix socket: Os { code: 30, kind: ReadOnlyFilesystem, message: "Read-only file system" } error](https://substrate.stackexchange.com/questions/8071/cannot-bind-unix-socket-os-code-30-kind-readonlyfilesystem-message-read/8116#8116).

In short -  relay chain is unable to spawn PVF worker when 'node.customNodeKey' option is defined. The reason is that custom node key is mounted into '/tmp/' folder and even more restricted to 'read-only' status.

Solution proposed: use custom paths mounted to avoid interference with generic folders, such as '/tmp' etc.

### DIFF
--- a/charts/node/Chart.yaml
+++ b/charts/node/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: node
 description: A Helm chart to deploy Substrate/Polkadot nodes
 type: application
-version: 4.4.1
+version: 4.4.2
 maintainers:
   - name: Parity
     url: https://github.com/paritytech/helm-charts


PierreBesson:
Thank you for the contribution @okalenyk !
This is a good solution. Using `/tmp` was not the wisest choice. @okalenyk can you please bump the patch version here https://github.com/paritytech/helm-charts/blob/main/charts/node/Chart.yaml

okalenyk:
> Thank you for the contribution @okalenyk !
> This is a good solution. Using `/tmp` was not the wisest choice. @okalenyk can you please bump the patch version here: https://github.com/paritytech/helm-charts/blob/main/charts/node/Chart.yaml#L5

Sure, @PierreBesson! Done. Thanks for reviewing.

PierreBesson (approved):

PR #239, opened 2023-05-05
Remove legacy rpc options

BulatSaif:
In this PR https://github.com/paritytech/substrate/pull/13384 the legacy rpc options was removed. 
## Changes

1. add support for the  legacy rpc options


BulatSaif:
blocked by: https://github.com/paritytech/cumulus/issues/2529

charts/node/README.md
```diff
@@ -70,6 +70,25 @@ node:
...
+--ws-external (replaced by --rpc-external)
+--unsafe-ws--external (replaced by --unsafe-rpc-external)
+--ipc-path (removed)
+--ws-port (replaced by --rpc-port)
+--ws-max-connections (replaced by --rpc-max-connections)
+--rpc-http (replaced by --rpc-addr)
+--rpc-ws (replaced by --rpc-addr)
+```
+New value was added to support this change:
+- `node.legacyRpcFlags`
```
PierreBesson, line 90: Add the following comment, if your node is still using the old RPC flags, please set `node.legacyRpcFlags=true`

charts/node/templates/service.yaml
```diff
@@ -75,8 +77,10 @@ spec:
     {{- $selectorLabels | nindent 4 }}
     statefulset.kubernetes.io/pod-name: {{ $fullname }}-{{ $i }}
   ports:
+    {{- if $.Values.node.legacyRpcFlags }}
     - port: {{ $.Values.node.perNodeServices.apiService.httpPort | int }}
       name: http-rpc
+    {{- end }}
     - port: {{ $.Values.node.perNodeServices.apiService.wsPort | int }}
       name: websocket-rpc
```
PierreBesson, line 85: Rename this port to `rpc`

charts/node/templates/statefulset.yaml
```diff
@@ -681,7 +685,11 @@ spec:
             periodSeconds: 10
             httpGet:
               path: /health
+              {{- if .Values.node.legacyRpcFlags }}
               port: http-rpc
+              {{- else }}
+              port: http-ws
```
PierreBesson, line 691: Change to `rpc`

charts/node/templates/statefulset.yaml
```diff
@@ -548,7 +548,9 @@ spec:
                 It is not a security risk unless they are exposed publicly.
                 */}}
                 --unsafe-rpc-external \
```
PierreBesson, line 550: There will be a problem because we always setting `--unsafe-rpc-external`, it was not a problem when this was only affecting the "internal" http port but we can't set it on the ws port for RPC nodes.

charts/node/templates/service.yaml
BulatSaif: we should keep old name in case someone uses ingress.

charts/node/templates/statefulset.yaml
BulatSaif: fixed

PierreBesson (approved):

PR #241, opened 2023-05-08
New rpc flags backward compatibility improvement

PierreBesson:
- Introduce new `rpc` named port
- Keep both `http-rpc` and `websocket-rpc` named port with deprecation warnings
- Replace hardcoded `9944` value in templates by `node.perNodeServices.apiService.rpcPort`


charts/node/templates/statefulset.yaml
```diff
@@ -665,24 +662,23 @@ spec:
...
           ports:
-            {{- if .Values.node.legacyRpcFlags }}
-            - containerPort: 9933
+            {{- if $.Values.node.legacyRpcFlags }}
+            - containerPort: {{ $.Values.node.perNodeServices.apiService.httpPort | int }}
               name: http-rpc
-              protocol: TCP
-            {{- end }}
-            - containerPort: 9944
+            - containerPort: {{ $.Values.node.perNodeServices.apiService.wsPort | int }}
```
BulatSaif, line 668: This may break old deployments.  
If `legacyRpcFlags` is `true` the  `--rpc-port` `--ws-port` flags are not set. The node will be listen port  `9933` and `9944`. Setting `.apiService.wsPort` or `.apiService.httpPort` will not affect the node but will break the service.

PierreBesson: I'm reverting to hardcoded 9933/9944 for legacy rpc flags in https://github.com/paritytech/helm-charts/pull/241/commits/9e9f48f6a21594a0080161bdbe2202b1047db40d

charts/node/values.yaml
```diff
@@ -332,8 +332,11 @@ node:
       externalTrafficPolicy: Cluster
       type: ClusterIP # or NodePort, LoadBalancer
       annotations: {}
+      rpcPort: 9944
+      # deprecated, use rpcPort
       httpPort: 9933
-      wsPort: 9944
+      # deprecated, use rpcPort
+      wsPort: 9955
```
BulatSaif, line 339: Can we keep the old value? 
```suggestion
      wsPort: 9944
```
Otherwise it may break someone deployment.

PierreBesson: This doesn't work as you can't have the same port number twice on a Service object. Either wsPort or rpcPort have to be set to something different. Would yo uprefer if I set wsPort=9944, rpcPort=9955 ?

PierreBesson: I agree this solution is better and less surprising for migrating from `websocket-rpc=944` to `rpc=9944`. I implemented it in https://github.com/paritytech/helm-charts/pull/241/commits/9e9f48f6a21594a0080161bdbe2202b1047db40d . I hardcode ports for the legacyRpcFlags=true but switch to variabilized ports (wsPort=9955 as default) for legacyRpcFlags=false

BulatSaif: Thank you, now  we will not break any old deployment if `legacyRpcFlags` is `true`

PR #243, opened 2023-05-11
Increase introspector's verbosity level

AndreiEres:


PierreBesson:
Please bump the version number in Chart.yaml

charts/polkadot-introspector/templates/deployment.yaml
```diff
@@ -52,6 +52,7 @@ spec:
               {{- end }}
               {{- end }}
               --ws {{ join "," .Values.introspector.rpcNodes }} \
+              -v \
```
lazam, line 55: Let's add a condition here to allow us to disable and enable verbose logging via values.yaml. We can have it as:

```
introspector.verboseLogging: false
```

Ideally, it should be false by default then we will just override it in our own value file.

lazam (approved):

PR #244, opened 2023-05-19
add new polkadot-stps chart

PierreBesson:
Add a new chart for https://github.com/paritytech/polkadot-stps to deploy a benchmark infrastructure.


bredamatt (changes requested):
Looks fine to me, but I noticed there were some GHA failures, so I guess as soon as those are fixed it should be ready.

PR #246, opened 2023-05-22
Allow configuring environment variables of WS Health exporter

bakhtin:
Add `wsHealthExporter.env` as a new variable to allow configuring [WS Health Exporter](https://github.com/paritytech/scripts/tree/master/dockerfiles/ws-health-exporter)


charts/node/templates/statefulset.yaml
```diff
@@ -780,6 +780,10 @@ spec:
           env:
             - name: NODE_URL
               value: ws://127.0.0.1:{{ .Values.node.perNodeServices.apiService.rpcPort | int }}
+            {{- range $key, $val := .Values.wsHealthExporter.env }}
+            - name: {{ $key }}
```
kogeler, line 784: Can we skip the host variable here?

charts/node/values.yaml
```diff
@@ -479,6 +479,7 @@ wsHealthExporter:
     repository: paritytech/ws-health-exporter
     tag: latest
```
kogeler, line 480:
```suggestion
    tag: 0a2e6e9b-20230412
```

Issue #176, opened 2022-10-10
Support externalized relay-chain node for collators


PierreBesson:
The parachain collator 0.9.300 supports [collation via RPC relay chain node](https://github.com/paritytech/cumulus/issues/989). In this mode the collator doesn't need to run a local relay-chain node and simply needs to point to a relay-chain RPC URL.

Add support for this mode:
- add `node.collatorRelayChain.rpcUrl` which when set will add `--relay-chain-rpc-url ws://rpc-node-url` and disable the relay-chain data and keystore volume
- remove collator flags when this is on

PierreBesson:
After experimenting a bit with external relay-chain node collation, we realize that it opens up new operational concerns:
- If the relay-chain node is inaccessible, the collator will fail fast.
- If we would run the relay-chain and para-chain node in 2 separate containers of the same pod, the relay-chain needs to be fully started before launching the para chain 
- Support for additional relay-chain node as failover if the default one is unavailable

PierreBesson:
Multiple RPC URLs are supported by passing a comma separated list
```
      --relay-chain-rpc-urls [<RELAY_CHAIN_RPC_URLS>...]
          EXPERIMENTAL: Specify an URL to a relay chain full node to communicate with
```

lazam:
https://github.com/paritytech/helm-charts/pull/251

PR #251, opened 2023-05-31
External Relay Chain Support

lazam:
https://github.com/paritytech/helm-charts/issues/176

This will add two new  values:
- node.externalRelayChain.enabled
- node.externalRelayChain.relayChainRpcUrls

Enabling it will only run the binary specific to parachain, and removes the related resources in running a collator with relay chain


charts/node/values.yaml
```diff
@@ -170,6 +170,20 @@ node:
...
+  ## Run the collator node without a relay chain
+  ## ref: https://github.com/paritytech/cumulus#external-relay-chain-node
+  ##
+  ## Enabling this option will disable the values of collatorRelayChain
+  ##
+  externalRelayChain:
+    enabled: false
+
+    ## List of Relay Chain RPCs to connect
+    relayChainRpcUrls: ["wss://rpc.polkadot.io"]
```
PierreBesson, line 184: I think it's better not to put a default value here (default:`[]`).

lazam: The reason why I added the polkadot RPC is because the default chain value is set to Polkadot. I can remove it.

charts/node/templates/statefulset.yaml
```diff
@@ -625,9 +633,6 @@ spec:
                 {{- if .Values.node.collatorRelayChain.customChainspecUrl}}
                 --chain={{ .Values.node.collatorRelayChain.customChainspecPath}} \
                 {{- end }}
-                {{- range .Values.node.telemetryUrls }}
-                --telemetry-url={{ . | squote }} \
-                {{- end }}
```
BulatSaif, line 630: why you removed `--telemetry-url`?

PR #253, opened 2023-06-05
Allow configuring requests/limits on node sidecars and staking-miner

bakhtin:
Allow configuring requests/limits on node sidecars and staking-miner


charts/node/values.yaml
```diff
@@ -54,6 +54,29 @@ googleCloudSdk:
   gsutilFlags: '-m -o "GSUtil:parallel_process_count=$(nproc --all)" -o "GSUtil:parallel_thread_count=2" -o "GSUtil:use_gcloud_storage=True"'
   # serviceAccountKey: ""
 
+## Sidecar containers
```
BulatSaif, line 57:
```suggestion
## initContainers
```

BulatSaif: are you sure you need resources on `initContainers`?

bakhtin: yes. This is required for policy enforcement tools like Gatekeeper to make sure that all Pods have resources configured. Including ephemeral ones like init containers

BulatSaif (approved):

PR #257, opened 2023-06-19
Reworking faucet configs: part 1

mutantcornholio:
This is part 1 of reworking faucet configs, where we moving static
network info back into code, so we would update it more easily.
However, we don't want to shoot ourselves at the foot, so we're
deploying it in three stages:
1. Add new variables to the configuration
2. Deploy the code (if delployment fails, we can safely rollback and
reiterate)
3. Remove old variables from the configuration

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -64,11 +64,15 @@ jobs:
 
     - name: Replace secrets in Helm charts
       env:
+        SMF_CONFIG_MATRIX_ACCESS_TOKEN: ${{ secrets.SMF_CONFIG_MATRIX_ACCESS_TOKEN }}
+        # TODO: remove these SMF_BACKEND_* / SMF_BOT_* vars after releasing corresponding faucet code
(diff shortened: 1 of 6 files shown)


mutantcornholio:
If I'm understanding the error in the tests, it's only about [this action](https://github.com/paritytech/helm-charts/blob/e336f0a4fbc234e074ce84ce669cb8e085518ac0/.github/workflows/pr.yaml#L67-L69) and the secrets to that are set in GitHub.  
In order not to update them twice, I can point the new vars in config to the old ones in secrets, and we'll update it with the part 3. 

We still need to something like that for secrets in vault, though...

BulatSaif:
You need to bump helm chart version in `charts/substrate-faucet/Chart.yaml`

mutantcornholio:
> You need to bump helm chart version in `charts/substrate-faucet/Chart.yaml`

Thanks, done

BulatSaif (approved):

Issue #237, opened 2023-04-26
Implement S3 support for backups


aardbol:
It could e.g. be achieved with one tool for both clouds: [s5cmd](https://hub.docker.com/r/peakcom/s5cmd), which potentially also downloads the backup faster than `gsutil`. It would be interesting to do some tests.

PR #262, opened 2023-06-29
Migrate to rclone and add helm-docs

bakhtin:
Migrate node snapshot restoration scripts to [rclone](https://rclone.org/). rclone is compatible with numerous cloud storage systems and allows to work with them using a unified syntax.

rclone replaces multiple different utilities that we have been using to download chain snapshots: `wget` for HTTP, `gcs` for Google Cloud Storage (GCS).

Add support for other providers -- for now just s3-compatible storages but can be easily expanded to support more. 

Add support for downloading snapshots over HTTP using a file with a list of files. It should significantly reduce cost of restoring a snapshot from S3/GCS as no expensive API requests will be consumed.

Bump major version of the chart introducing breaking changes (see README)

**Bonus**: Add `helm-docs` to auto-generate values section of README (see README.md and README.md.gotmpl)

TODO:
- ~~Update `Deploying a node with data synced from a snapshot archive` section of README~~
- ~~Test different scenarios for backup restoration~~

Closes: #237


dblane-digicatapult (commented):
Excellent set of changes.

charts/node/values.yaml
```diff
@@ -316,176 +370,201 @@ node:
...
 
+  # -- Annotations to add to the Service
   serviceAnnotations: {}
 
-  ## Use deprecated ws/rpc flags.
-  ## ref: https://github.com/paritytech/substrate/pull/13384
-  ##
+  # -- Use deprecated ws/rpc flags.
+  # ref: https://github.com/paritytech/substrate/pull/13384
   legacyRpcFlags: false
```
BulatSaif, line 420: This is good time to break rpc port backward comparability. We need to remove option `legacyRpcFlags`, cleanup all `httpPort` and `wsPort` and keep only `rpcPort`.

bakhtin: Parity still runs some archive nodes on v0.9.42. Removing the flag will break the installations. I'd suggest to wait before v0.9.44 is released and we upgraded all the nodes to v0.9.43.

charts/node/values.yaml
```diff
@@ -1,272 +1,333 @@
...
+    # -- (string) Path on the volume to store chain data
+    chainPath:
+    # -- Storage class to use for persistent volume
     storageClass: ""
+    # -- Size of the volume for chain data
     volumeSize: 100Gi
+    # -- Annotations to add to the volumeClaimTemplates
     annotations: {}
 
+    # -- Configure parameters for restoring chain snapshot. Uses [rclone](https://rclone.org/)
```
PierreBesson, line 199: Truly excellent design of the snapshot configuration. Could you also add a variable for passing custom flags to rclone in case this is needed ?

PierreBesson: Never mind I see this is already possible in the `downloadChainSnapshot` block.

charts/node/values.yaml
```diff
@@ -1,272 +1,333 @@
...
+    # -- Adds `-x` shell option to container.
+    # Note: passwords and keys used in container may appear in logs
+    debug: false
+    # -- Flags to add to the CLI command. We rely on rclone for downloading snapshots so make sure the flags are compatible.
+    cmdArgs: ""
+
+  downloadChainspec:
+    image:
+      # -- Image repository
+      repository: paritytech/kubetools-kubectl
```
PierreBesson, line 57: For the downloadChainspec and downloadRuntime init container as we are only using `sh` and `wget`, it should be possible to only use `alpine:latest`.

PierreBesson (changes requested):

charts/node/values.yaml
bakhtin: it is there: `initContainers.downloadChainSnapshot.cmdArgs`

charts/node/templates/statefulset.yaml
```diff
@@ -79,183 +79,150 @@ spec:
...
+
+                elif [ "${METHOD}" == "http-single-tar" ]; then
+                  rclone copyurl {{ .Values.initContainers.downloadChainSnapshot.cmdArgs }} --stdout --error-on-no-transfer ${SNAPSHOT_URL} | tar -x -C /chain-data/chains/${RELAY_CHAIN_PATH}/
+
+                elif [ "${METHOD}" == "gcs" ]; then
+                  LATEST=$(rclone cat {{ .Values.initContainers.downloadChainSnapshot.cmdArgs }} --quiet :gcs:${SNAPSHOT_URL}/latest_version.meta.txt)
+                  if [ -z "$LATEST" ]; then
+                    echo "Failed to retrieve latest_version.meta.txt file. Will download everything from ${SNAPSHOT_URL} instead"
+                  fi
+                  rclone sync {{ .Values.initContainers.downloadChainSnapshot.cmdArgs }} --fast-list --transfers $PARALLEL_TRANFERS --progress --error-on-no-transfer :gcs:${SNAPSHOT_URL}/${LATEST} /chain-data/chains/${RELAY_CHAIN_PATH}/{{ $databasePath }}/
```
kogeler, line 179: Does it work with public buckets without the [option](https://rclone.org/googlecloudstorage/#gcs-bucket-policy-only). I mean buckets that use Uniform bucket-level access.

charts/node/templates/statefulset.yaml
```diff
@@ -79,183 +79,150 @@ spec:
...
+
+                elif [ "${METHOD}" == "http-filelist" ]; then
+                  LATEST=$(rclone copyurl {{ .Values.initContainers.downloadChainSnapshot.cmdArgs }} --stdout ${SNAPSHOT_URL}/latest_version.meta.txt )
+                  if [ -z "$LATEST" ]; then
+                    echo "Failed to retrieve latest_version.meta.txt file. Will download everything from ${SNAPSHOT_URL} instead"
+                  else
+                    SNAPSHOT_URL="${SNAPSHOT_URL}/${LATEST}"
+                  fi
+                  rclone copyurl {{ .Values.initContainers.downloadChainSnapshot.cmdArgs }} --error-on-no-transfer ${SNAPSHOT_URL}/{{ .Values.node.chainData.chainSnapshot.filelistName }} /tmp/filelist.txt
+                  rclone copy {{ .Values.initContainers.downloadChainSnapshot.cmdArgs }} --progress --error-on-no-transfer --transfers $PARALLEL_TRANFERS --http-url ${SNAPSHOT_URL} --no-traverse --http-no-head --disable-http2 --files-from /tmp/filelist.txt :http: /chain-data/chains/${CHAIN_PATH}/{{ $databasePath }}/
```
kogeler: Did you check how the results of the `--progress`  flags are stored in stdout logs? It uses terminal string rewriting. You can replace it with the `-v` flags to have full logs that contain information about all downloaded files.

charts/node/README.md.gotmpl
```diff
@@ -0,0 +1,208 @@
+<!--
```
kogeler, line 1: Add links to http backups from `snapshots.polkadot.io`, please.

bakhtin: done

charts/node/templates/statefulset.yaml
bakhtin: Well, it looks ugly with `--progress`. But it shows restoration speed, actual progress in percentage, and files restored . With `-v` AND `--progress` it will be even more uglier due to the mentioned string rewriting. I'd rather keep only `--progress` flag as I see more befit in having live stats over list of restored files printed to the console.

kogeler (approved):

PR #267, opened 2023-07-04
Add cleanup step on failed chain snapshot restoration

bakhtin:
Chain snapshot restoration script may fail leaving empty chain data directory on the disk. 

On the first run of the script the chain data directory is created. If the script fails the directory remains empty while K8s will re-run failed init container. On the second execution the script will detect that chain data directory is present and will skip snapshot restoration even though it was not successful in the first try. The empty directory will then be mounted to the main node container and the chain will start syncing from scratch (possibly taking many days).

Add a `trap` command to do a cleanup (removing the chain directory) if any error occurs during the snapshot restoration process and avoid above situation from happening.

### DIFF
--- a/charts/node/templates/statefulset.yaml
+++ b/charts/node/templates/statefulset.yaml
@@ -87,10 +87,10 @@ spec:
             - -c
             - |
               set -eu -o pipefail {{ if .Values.initContainers.downloadChainSnapshot.debug }}-x{{ end }}
-              trap 'echo -e "Snapshot restoration failed. Checkout the logs for errors.\nRemoving /chain-data/chains/${CHAIN_PATH}/{{ $databasePath }} ..."; rm -rf /chain-data/chains/${CHAIN_PATH}/{{ $databasePath }}' ERR
               if [ -d "/chain-data/chains/${CHAIN_PATH}/{{ $databasePath }}" ]; then
                 echo "Database directory already exists, skipping chain snapshot download"
               else
+                trap 'echo -e "Snapshot restoration failed. Checkout the logs for errors.\nRemoving /chain-data/chains/${CHAIN_PATH}/{{ $databasePath }} ..."; rm -rf /chain-data/chains/${CHAIN_PATH}/{{ $databasePath }}' ERR
                 PARALLEL_TRANFERS="$(($(nproc --all) * 5 < 50 ? $(nproc --all) * 5 : 50))" # MIN(vCPU_count * 5, 50)
                 echo "Downloading chain snapshot"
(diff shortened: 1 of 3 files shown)


BulatSaif (approved):

charts/node/templates/statefulset.yaml
```diff
@@ -86,6 +86,7 @@ spec:
           args:
             - -c
             - |
+              trap 'echo -e "Snapshot restoration failed. Checkout the logs for errors.\nRemoving /chain-data/chains/${CHAIN_PATH}/{{ $databasePath }} ..."; rm -rf /chain-data/chains/${CHAIN_PATH}/{{ $databasePath }}' ERR
```
BulatSaif, line 89: can you move this after  this if condition?
`if [ -d "/chain-data/chains/${CHAIN_PATH}/{{ $databasePath }}" ];`

BulatSaif: for example we update base container to alpine and now `set -eu -o pipefail` option is not supported. This trap may remove working DB.

BulatSaif (changes requested):
This script executed on each restart we should make sure that trap will not remove working db

charts/node/templates/statefulset.yaml
PierreBesson: Let's make sure this can never be executed on a node which already has a populated disk.

bakhtin: > can you move trap after this if condition?
if [ -d "/chain-data/chains/${CHAIN_PATH}/{{ $databasePath }}" ];

Moving it inside `if` won't make much sense as it will define `trap` instruction only if directory already exists. Any errors in the `else` scope will not be trapped and this is where the main logic of the script is happening.

I can move `trap` right after `set` instruction so that if `set` is using incompatible options `trap` will not execute it's cleanup logic.

The script still checks for chain data directory existence before going any further. So you just need the first `if` to run without errors for `trap` to not be executed:
```
if [ -d "/chain-data/chains/${CHAIN_PATH}/{{ $databasePath }}" ]; then
  echo "Database directory already exists, skipping chain snapshot download"
```

btw, `set -eu -o pipefail` IS supported in Alpine images

BulatSaif: > Any errors in the else scope will not be trapped and this is where the main logic of the script is happening.

Yes, my proposal was to move it inside else block, sorry for miscommunication.  Also can you do it for `download-relay-chain-snapshot` init container too?

BulatSaif (approved):

PR #278, opened 2023-07-12
Remove faucet SMF_CONFIG_DEPLOYED_TIME env var

PierreBesson:
This is not a very useful feature to have and it is causing constant redeploys in our ArgoCD deployment every time our git config is has a new commit.


mutantcornholio (approved):
Ready to merge

PR #286, opened 2023-08-17
Added cloudsql to faucet

mutantcornholio:
Couldn't find an existing chart here with cloudsql enabled, so I'm not really sure if this is enough on the `helm-charts` side. 

A part of https://github.com/paritytech/polkadot-testnet-faucet/issues/335

### DIFF
--- a/charts/substrate-faucet/values.yaml
+++ b/charts/substrate-faucet/values.yaml
@@ -125,3 +125,20 @@ ingress:
   ## @param ingress.enabled Specifies whether as Ingress should be created
   ##
   enabled: false
+
+postgresql:
+  global:
+    postgresql:
+      auth:
+        postgresPassword: "Secret!"
+  primary:
+    persistence:
+      storageClass: "ssd-csi"
+      size: 4Gi
+    resources:
+      limits:
+        cpu: 500m
+        memory: 1024Mi
+      requests:
+        cpu: 250m
+        memory: 512Mi
--- a/charts/substrate-faucet/values.yaml
+++ b/charts/substrate-faucet/values.yaml
@@ -133,7 +133,6 @@ postgresql:
         postgresPassword: "Secret!"
   primary:
     persistence:
-      storageClass: "ssd-csi"
       size: 4Gi
     resources:
       limits:
--- a/charts/substrate-faucet/Chart.yaml
+++ b/charts/substrate-faucet/Chart.yaml
@@ -2,7 +2,11 @@ apiVersion: v2
 name: substrate-faucet
 description: A Helm chart to deploy substrate-faucet
 type: application
-version: 2.2.2
+version: 3.0.0
 maintainers:
   - name: Parity
     url: https://github.com/paritytech/helm-charts
+dependencies:
+  - name: postgresql
+    version: "12.8.3"
+    repository: "https://charts.bitnami.com/bitnami"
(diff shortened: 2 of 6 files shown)


mutantcornholio:
hmmm, where are the secrets for the integration tests stored? Were they ever passing?

BulatSaif:
> hmmm, where are the secrets for the integration tests stored? Were they ever passing?

Secrets are defined here: https://github.com/paritytech/helm-charts/blob/d7a6b48348368cbe2fae32a8529954a4c6a67392/charts/substrate-faucet/ci/kind-values.yaml#L11

I think the token was rotated, and we did not update it in github:
image


I am looking for a password for `@parity_ci_bot:matrix.org` to generate new token.

charts/substrate-faucet/values.yaml
```diff
@@ -125,3 +125,10 @@ ingress:
   ## @param ingress.enabled Specifies whether as Ingress should be created
   ##
   enabled: false
+
+## Database configuration
+##
+cloudsql:
```
BulatSaif, line 131: What kind of data will be stored in SQL? Do you really need CloudSQL, or will just a PostgreSQL container be enough? 
If you don't require backups or a multi-replica database, running PostgreSQL in Kubernetes is preferred since it is easier to maintain and more cost-efficient.

mutantcornholio: Good point. We can easily survive data loss here. 

It'll be used for two things: 
* store historical "drips", to ensure daily drip limits (current functionality)
* as an AMQP-substitute (using pg-boss library). Faucet will be running multiple instances, all of which will produce and consume jobs.

Is there an a chart with postgres in k8s, that I could use as an example?

BulatSaif: > Is there an a chart with PostgreSQL in k8s, that I could use as an example?

One of the popular Helm charts for PostgreSQL is the Bitnami chart. Here's an example of how to use it:

1. To the end of the file `charts/substrate-faucet/Chart.yaml`, add the following:

```yaml
dependencies:
  - name: postgresql
    version: "12.8.3" 
    repository: "https://charts.bitnami.com/bitnami"
```

2. In the `values.yaml` file, add the following configuration:

```yaml
postgresql:
  global:
    postgresql:
      auth:
        postgresPassword: "Secret!"
  primary:
    persistence:
      storageClass: "ssd-csi"
      size: 20Gi
    resources:
      limits:
        cpu: 1000m
        memory: 2048Mi
      requests:
        cpu: 500m
        memory: 1024Mi
```

You can find more available parameters [here](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml). Note that these parameters should be placed under the `postgresql` root key, which is the name of the dependency.

BulatSaif:
@mutantcornholio,  thank you for your contribution. You did an excellent job with the PostgreSQL deployment. However, it's not entirely clear to me how the application will establish a connection to the database.

I've pushed a commit and added an environment variable to the application container containing the PostgreSQL credentials. You can verify that the database is functioning correctly from the application container using the following command:
```bash
apk add postgresql-client
PGPASSWORD=$POSTGRESQL_PASSWORD psql --username=$POSTGRESQL_USERNAME --host=$POSTGRESQL_HOST --port=$POSTGRESQL_PORT --dbname=$POSTGRESQL_DATABASE
```

I've used the `POSTGRESQL_` prefix for the environment variable. If your application expects different environment variables, please feel free to update the names accordingly.

BulatSaif (approved):

mutantcornholio:
> I've pushed a commit and added an environment variable to the application container containing the PostgreSQL credentials. You can verify that the database is functioning correctly from the application container using the following command:

tbh, I was thinking on passing it on the usage side, in `cloud-infra`. But now when I think of it, that'd be definitely worse :)

> I've used the POSTGRESQL_ prefix for the environment variable. If your application expects different environment variables, please feel free to update the names accordingly.

Yeah, it does expect those, alinged with bitnami images.

PR #289, opened 2023-08-28
Remove runtime from pokadot-introspector

AndreiEres:
Closes https://github.com/paritytech/devops/issues/2827


lazam:
Please bump the version of the chart here: https://github.com/paritytech/helm-charts/blob/main/charts/polkadot-introspector/Chart.yaml

PierreBesson (approved):

PierreBesson:
@AndreiEres it looks like the new docker image doesn't have the binary available at the expected path, hence the failing build.

> /bin/sh: 1: exec: polkadot-introspector/polkadot-block-time: not found

AndreiEres:
> it looks like the new docker image doesn't have the binary available at the expected path, hence the failing build.

That's true, I haven't released new images to not break all our existing deployments for a long time.

PR #295, opened 2023-08-30
[Testnet-Manager] Add role in serviceaccount to read volumes

lazam:
Create two new permissions:

* volume read acces  to query volumes used by pods
* node read access to query metadata where the pod is running

Needed for https://github.com/paritytech/testnet-manager/pull/99

### DIFF
--- a/charts/testnet-manager/templates/rbac.yaml
+++ b/charts/testnet-manager/templates/rbac.yaml
@@ -38,7 +38,7 @@ rules:
 ---
 # Creat Role bindings
 {{- $serviceAccountName := include "testnet-manager.serviceAccountName" . -}}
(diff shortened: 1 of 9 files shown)


charts/testnet-manager/templates/rbac.yaml
```diff
@@ -74,3 +83,18 @@ subjects:
     name: {{ $serviceAccountName }}
     namespace: {{ $.Release.Namespace }}
 ---
+# Allow the testnet-manager service account to read volumes in the rococo namespace
```
BulatSaif, line 86:
```suggestion
# Allow the testnet-manager service account to read volumes in the current namespace
```

BulatSaif (approved):

BulatSaif (approved):

PR #297, opened 2023-09-04
Expose the api-sidecar prometheus port on the service

PierreBesson:
This fixes prometheus scraping set on the service-monitor.


BulatSaif (approved):
Please, fix CI before merging.

PR #300, opened 2023-09-06
Add support to run new feature: light client mode

lazam:
To support a new feature of Substrate to run the node as light client mode in this [PR](https://github.com/paritytech/cumulus/pull/2270)


bakhtin:
Please, make sure to edit `README.md.gotmpl` instead of directly editing `README.md`, otherwise `pre-commit` hooks will be failing. We recently added [Contributing guide](https://github.com/paritytech/helm-charts/blob/main/CONTRIBUTING.md) which demonstrates how to setup a local environment to work with `pre-commit`.

charts/node/templates/statefulset.yaml
```diff
@@ -604,6 +607,10 @@ spec:
                 {{- end }}
                 {{- if and .Values.node.isParachain (not .Values.node.externalRelayChain.enabled) }}
                 -- \
+                {{- if  .Values.node.lightClientMode.enabled }}
+                --chain={{ if .Values.node.lightClientMode.relayChainCustomChainspecUrl }}{{ .Values.node.lightClientMode.relayChainCustomChainspecPath }}{{ else }}{{.Values.node.lightClientMode.relayChain}}{{ end }}
```
BulatSaif, line 611: we already defined `--chain=` flag twice (before and after `--`), defining it third time may introduce some confusions. 

I believe you can use just this:
```
{{ if .Values.node.lightClientMode.relayChainCustomChainspecUrl }}
--chain={{ .Values.node.lightClientMode.relayChainCustomChainspecPath }}
{{ end }}
```

If chain is not defined the substrate will take value from parachain chainspec [relay_chain](  https://github.com/paritytech/chainspecs/blob/main/rococo/parachain/rockmine2/chainspec-nonraw.json#L18 ) key

charts/node/values.yaml
```diff
@@ -248,12 +248,29 @@ node:
...
+    relayChainCustomChainspecUrl: ""
+
   collatorRelayChain:
     # -- Name of the Relay Chain to connect
     chain: polkadot
 
     # -- Path to the file containing the chainspec of the collator relay-chain
-    customChainspecPath: "/relaychain-data/relay_chain_chainspec.json"
+    # Set to /relaychain-data to use additional volume
+    customChainspecPath: ""
```
BulatSaif, line 273: Changing default value may break all current deployment.

charts/node/values.yaml
```diff
@@ -248,12 +248,29 @@ node:
     # -- List of Relay Chain RPCs to connect
     relayChainRpcUrls: []
 
+  # -- EXPERIMENTAL!!!
+  # Run the collator node without a relay chain via light client
+  # ref: https://github.com/paritytech/cumulus/pull/2270
+  # Enabling this option will disable the values of collatorRelayChain
+  lightClientMode:
```
BulatSaif, line 255: Can you enable both the `externalRelayChain` and `lightClientMode` features together?

If they are mutually exclusive, I recommend linking the `externalRelayChain` and `lightClientMode` features together, as they serve the same purpose:

```yaml
{{ if externalRelayChain.enabled }} 
  {{ if externalRelayChain.relayChainRpcUrls }}
    --relay-chain-rpc-urls {{- range externalRelayChain.relayChainRpcUrls }} "{{ . }}" {{- end }} \                
  {{ else }}
    --relay-chain-light-client \ 
  {{ end }}
{{ end }}
```

charts/node/values.yaml
lazam: Yes. I just realized that users might not add it in their helm chart and breaks all deployment. I reverted it.

charts/node/values.yaml
lazam: I added a different logic here. The `externalRelaychain`, `lightClient` and even the regular collator with relaychain are just different modes to run a collator.

charts/node/templates/statefulset.yaml
lazam: We should retain the additional `--chain` for now since we're not sure if it has the same behavior of a regular collator mode since it uses light client.

charts/node/values.yaml
```diff
@@ -239,20 +239,35 @@ node:
...
+  # Enabling this option will disable the values of collatorRelayChain
+  collatorLightClient:
+    # -- Enable deployment of the external collator
+    enabled: false
+    # -- Name of the Relay Chain to connect
+    relayChain: ""
+    # -- Path to the file containing the chainspec of the collator relay-chain
+    relayChainCustomChainspecPath: "/chain-data/relay_chain_chainspec.json"
+    # -- (string) URL to retrive custom chain spec
+    relayChainCustomChainspecUrl:
```
BulatSaif, line 263: `relayChain` and `relayChainCustomChainspecUrl` is already present:  `collatorRelayChain.chain`  `collatorRelayChain.customChainspecUrl`, it is very confusing without looking at the code why it is done like that. 
Since this is an experimental feature I don't mind having them duplicated. But for production usage, we should think of moving them under the `collatorRelayChain` section.:
```
.Values.node.collatorExternalRelayChain -> .Values.node.collatorRelayChain.externalRelayChain
.Values.node.collatorLightClient ->  .Values.node.collatorRelayChain.lightClient
```

charts/node/templates/statefulset.yaml
```diff
@@ -699,7 +715,7 @@ spec:
               name: chain-data
             - mountPath: /keystore
               name: chain-keystore
-            {{- if and .Values.node.isParachain (not .Values.node.externalRelayChain.enabled)}}
+            {{- if and .Values.node.isParachain (and (not .Values.node.collatorExternalRelayChain.enabled ) (not .Values.node.collatorLightClient.enabled)) }}
```
BulatSaif, line 718: you can avoid having nested `add`, the `add` function [supports](https://helm.sh/docs/chart_template_guide/function_list/#and) multiple arguments:
```suggestion
            {{- if and .Values.node.isParachain (not .Values.node.collatorExternalRelayChain.enabled ) (not .Values.node.collatorLightClient.enabled) }}
```

BulatSaif (approved):

charts/node/values.yaml
lazam: I disagree with moving them under `collatorRelayChain`. `collatoRelayChain` section is another mode to run a collator which runs both parachain and relaychain node in it while `collatorExternalRelayChain` and `collatorLightClient` are also another type to run a collator which is entirely independent to `collatorRelayChain`. 

My suggestion is to improve the naming and the fields we use. Something like:

```
collator:
  enabled: true
  
  relayChainNode:
    enabled: false
    All specific fields for relaychain node

  lightClientNode:
    enabled: false

  externalRelayChain:
     enabled: false
     rpcUrls: []
     
     
  chainSpec: xxx
  other shared fields: xxx
```

BulatSaif: > My suggestion is to improve the naming and the fields we use. Something like:
> ```
> collator:
>   enabled: true
>   
>   relayChainNode:
>     enabled: false
> ``` 

Now I see your point. Yes, your proposed mapping is better.  But since it will require a bumping a major version we can also rename the remaining fields. Something like:  

```
parachain:
  enabled: true
  
  embeddedRelayChainNode:
    enabled: false
    All specific fields for relaychain node

  lightClientRelayChainNode:
    enabled: false

  rpcRelayChainNode:
     enabled: false
     rpcUrls: []
     
     
  chainSpec: xxx
  other shared fields: xxx
```

PierreBesson: I agree that this would be a clearer config format (especially with @BulatSaif formulation). In the past we couldn't have anticipated that there would be so many "relay-chain node modes" in collators. Please propose this in a separate issue so we can discuss it.

charts/node/README.md
```diff
@@ -96,6 +96,10 @@ node:
 
 ## Upgrade
 
+### From v5.x.x to v5.3.0 (⚠️ breaking changes)
```
PierreBesson, line 99: Make sure to state that those are breaking changes in an experimental option only. Otherwise this would require a major version bump. As it is, it's unlikely to affect anyone besides us.

lazam: Yes. I added two of the critical information. 

The version from which is affected starting from 5, and the variable that was changed.

charts/node/templates/statefulset.yaml
```diff
@@ -644,10 +660,10 @@ spec:
                 --listen-addr=/ip4/0.0.0.0/tcp/30334/ws \
                 {{- end }}
                 {{- end }}
-                {{- if  (not .Values.node.externalRelayChain.enabled) }}
+                {{- if  (and (not .Values.node.collatorExternalRelayChain.enabled) (not .Values.node.collatorLightClient.enabled)) }}
                 --listen-addr=/ip4/0.0.0.0/tcp/30333 \
                 {{- end }}
-
+                {{- end }}
```
BulatSaif, line 666: Move this  `{{- end }}` up:
Before:
```
--parachain flags 
{{- if and .Values.node.isParachain (not .Values.node.externalRelayChain.enabled) }}
-- \ 
--relaychain flags
{{- end }}
--common flags  (parachain+relaychain)
 ```
 
 Now:
 
```
--parachain flags 
{{- if and .Values.node.isParachain (not .Values.node.externalRelayChain.enabled) }}
-- \ 
--new relaychain flags
{{- if  not .Values.node.collatorLightClient.enabled }}
--relaychain flags
{{- end }}
--common flags  (parachain+relaychain)
{{- end }}
 ```
 
 expected:
 ```
--parachain flags 
{{- if and .Values.node.isParachain (not .Values.node.externalRelayChain.enabled) }}
-- \ 
--new relaychain flags
{{- if  not .Values.node.collatorLightClient.enabled }}
--relaychain flags
{{- end }}
{{- end }}
--common flags  (parachain+relaychain)
 ```

Issue #296, opened 2023-09-04
Expand test scenarios


bakhtin:
Each Helm chart is now configured to run a smoke test on each PR. A chart is deployed to a local K8s cluster with default or close-to-default parameters. While smoke tests are still good to capture basic errors, they don't cover real-world usage scenarios.

To prevent the bugs from ending up in the release more comprehensive end-to-end tests should be added to the CI pipeline. But these tests should not repeat or serve as a replacement for the tests of the application itself. Rather the tests should capture Helm chart specific errors like improperly exposed ports, failures to attach PVCs, bugs in the init-scripts, etc.

The `values.yaml` file should also be extended with more comprehensive configuration reflecting how the chart is used in the wild. Potentially, having multiple `values.yaml`s to cover different use-cases.

bakhtin:
Check out https://github.com/bitnami/charts/blob/main/TESTING.md

bakhtin:
As the first step I'll pick up node Helm chart to implement tests for it. Node Helm chart is the most used and complex chart in the repo. 

The initial iteration should cover the most used configuration options and the pieces of functionality that has the most impact on the runtime. I came up with the following:

Test the following node modes:
- Validator
- RPC
- Full
- Bootnode

Test scenarios:
- a node is running in `--dev` mode along with `ws-exporter`` sidecar and posting a healthy status (relay chain + para chain)
- backup restoration works. Need to mock it somehow, restoring full backup is not feasible. Possibly, serve a small backup for a custom chain from the local HTTP server spun up within the same test suite
- RPC node (2 replicas) is publicly exposed through the LB and serves RPC queries
- a Bootnode can bootstrap other nodes on the network
- a Validator nodes produce blocks (possible with `--dev` flag?)
- can load keys from Vault
- Pods are reachable through Services (p2p, rpc)
- can download custom chainspec
- can download custom runtime
- generated node key are persisted
- can inject keys

I found the [Bitnami's testing guide](https://github.com/bitnami/charts/blob/main/TESTING.md?rgh-link-date=2023-09-04T15%3A48%3A07Z) comprehensive covering the details of possible strategies and tools for testing Helm charts.  Bitnami's Helm charts [repo](https://github.com/bitnami/charts) uses 3 tools to run the tests:
[...]

PR #304, opened 2023-09-28
Expand test scenarios

bakhtin:
https://github.com/paritytech/helm-charts/issues/296#issuecomment-1713929440

Add Golang-based tests ([Ginkgo](https://onsi.github.io/ginkgo/#top)) for `node` Helm chart. Tests are running on each commit to an opened PR.
Each test is running against a chart deployed to a local K8s cluster (`kind`). Same chart may be deployed with different configuration using multiple values files.

Currently, the following tests are implemented:
- An RPC node (2 replicas) deployed behind the load balancer is posting a healthy status (checked with `ws-health-exporter` sidecar)
- A collator node is posting a healthy status after the deployment
- A collator node can successfully restore using the remote chain backup

More tests to come in the following PRs. The goal is to cover most common configurations of the chart with the tests.

All tests are stored in `ci` directory of the chart. A CI pipeline sets up the test suite following these major steps:
- Set up local K8s cluster (using kind)
- Deploy addons. Currently `metallb` (bare-metal load balancer), `nginx-ingress-controller`, `metrics-server`, and `nginx-assets-server` (for serving static content over HTTP, e.g., chainspec)
- Deploy various configurations of the chart (using multiple values files)
- Run Ginkgo tests against the live version of the deployment

I also updated the port name in the Service as K8s rejects names longer than 15 chars. Not sure how it worked before :shrug: 


.github/workflows/pr.yaml
```diff
@@ -84,12 +85,21 @@ jobs:
...
 
-    - name: Run chart-testing (install)
+    - uses: actions/setup-go@v4
+      with:
+        go-version: '1.20'
+        cache-dependency-path: "charts/**/ginkgo/go.sum"
+
+    - name: Run ginkgo tests
       run: |
-        ct install \
```
PierreBesson, line 89: Personally, I would not remove this completely but just for the node helm-chart. This is quite useful for having basic chart tests on the other helm-charts.

PierreBesson (changes requested):

.github/workflows/pr.yaml
bakhtin: My bad, I removed it for testing and forgot to rollback the change. I reverted it excluding `node` chart as it is not covered by proper tests.

PierreBesson (approved):

Issue #233, opened 2023-04-05
Replace --pruning by --state-pruning in node helm chart


PierreBesson:
Use the new flag name introduced in https://github.com/paritytech/substrate/pull/11934
The current `--pruning` still works but doesn't show up in `polkadot --help` anymore so let's use the new one.

PR #312, opened 2023-11-15
[Enhancement] Replace with the correct blocks pruning args

lazam:
fixes: https://github.com/paritytech/helm-charts/issues/233

The latest version of the sdk uses blocks-pruning instead of pruning. The old usage of deprecated args might result to unexpected behavior.

### DIFF
--- a/charts/node/README.md.gotmpl
+++ b/charts/node/README.md.gotmpl
@@ -107,6 +107,9 @@ node:
 ```
 
 ## Upgrade
+### From v5.x.x to v5.5.0 (⚠️ breaking changes)
+- The pruning flag is now using `--blocks-pruning` which starts from polkadot version 0.928
+- The flag `--pruning` is now allowed in both .Values.node.flags and .Values.node.collatorRelayChain.flags. When using `--pruning`, ensure that the values of .Values.node.chainData.pruning and .Values.node.collatorRelayChain.chainData.pruning are explicitly set to false to maintain previous behavior.
 
 ### From v5.x.x to v5.3.0 (⚠️ breaking changes)
 - The following flags have changed:
--- a/charts/node/README.md.gotmpl
+++ b/charts/node/README.md.gotmpl
@@ -108,7 +108,7 @@ node:
 
 ## Upgrade
 ### From v5.x.x to v5.5.0 (⚠️ breaking changes)
-- The pruning flag is now using `--blocks-pruning` which starts from polkadot version 0.928
+- The pruning flag is now using `--blocks-pruning` which starts from polkadot version v0.9.28
 - The flag `--pruning` is now allowed in both .Values.node.flags and .Values.node.collatorRelayChain.flags. When using `--pruning`, ensure that the values of .Values.node.chainData.pruning and .Values.node.collatorRelayChain.chainData.pruning are explicitly set to false to maintain previous behavior.
 
 ### From v5.x.x to v5.3.0 (⚠️ breaking changes)
--- a/charts/node/README.md.gotmpl
+++ b/charts/node/README.md.gotmpl
@@ -109,7 +109,7 @@ node:
 ## Upgrade
 ### From v5.x.x to v5.5.0 (⚠️ breaking changes)
 - The pruning flag is now using `--blocks-pruning` which starts from polkadot version v0.9.28
-- The flag `--pruning` is now allowed in both .Values.node.flags and .Values.node.collatorRelayChain.flags. When using `--pruning`, ensure that the values of .Values.node.chainData.pruning and .Values.node.collatorRelayChain.chainData.pruning are explicitly set to false to maintain previous behavior.
+- The flag `--pruning` is now allowed in both .Values.node.flags and .Values.node.collatorRelayChain.flags. When using `--pruning`, ensure that the values of .Values.node.chainData.pruning and .Values.node.collatorRelayChain.chainData.pruning are explicitly set to `false` to maintain previous behavior.
 
 ### From v5.x.x to v5.3.0 (⚠️ breaking changes)
 - The following flags have changed:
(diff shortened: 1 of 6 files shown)


charts/node/templates/_helpers.tpl
```diff
@@ -98,5 +98,5 @@ Create the database path depending on the database backend in use (rocksdb or pa
 Define a regex matcher to check if the passed node flags are managed by the chart already
 */}}
 {{- define "node.chartManagedFlagsRegex" -}}
-{{- "(\\W|^)(--name|--base-path|--chain|--validator|--collator|--light|--database|--pruning|--prometheus-external|--prometheus-port|--node-key|--wasm-runtime-overrides|--jaeger-agent|--rpc-external|--unsafe-rpc-external|--ws-external|--unsafe-ws-external|--rpc-methods|--rpc-cors|--rpc-port|--ws-port|--enable-offchain-indexing)(\\W|$)" }}
+{{- "(\\W|^)(--name|--base-path|--chain|--validator|--collator|--light|--database|--pruning|--blocks-pruning|--prometheus-external|--prometheus-port|--node-key|--wasm-runtime-overrides|--jaeger-agent|--rpc-external|--unsafe-rpc-external|--ws-external|--unsafe-ws-external|--rpc-methods|--rpc-cors|--rpc-port|--ws-port|--enable-offchain-indexing)(\\W|$)" }}
```
BulatSaif, line 101: If we need to deploy an old node or 3rd party node which do not support `--blocks-pruning` how we can do it with a new helm chart?

lazam: There's two ways that I can think of on how to solve this:

1. We should start using the proper appVersion value in the Chart. This approach allows us to determine which version the chart supports, and create certain conditions based from the appVersion. If we add this, it provides a visibility in which app version supports for specific version of the chart, the documentation and communication is clear just by using the chart, and it will be consistent through out different environment.

2. Another option is to just rely on specific node.image.tag condition for this. 

I prefer the first approach since it is how Helm chart should be written.

BulatSaif: 1. appVersion will work only for parity-managed clients, but you also have kagome, moonbase which has completely different versioning.

2. image tag can be anything e.g master

One of the solutions that I see is to allow users to set `--pruning` as custom flags, for example:
```
.Values.node.chainData.pruning = -1
 .Values.node.flags = ['--pruning=1000']

```
 this will do the trick, but currently both `--pruning, --blocks-pruning` is forbidden by `chartManagedFlagsRegex`.

lazam: Yeah that make sense. I forgot about other app implementation that uses polkadot-sdk. The solution is to allow passing of the pruning flags, but to also implement a check to disable the default pruning value if we pass the flag. This is to avoid modifying the pruning values in chart.

charts/node/README.md.gotmpl
```diff
@@ -107,6 +107,9 @@ node:
 ```
 
 ## Upgrade
+### From v5.x.x to v5.5.0 (⚠️ breaking changes)
+- The pruning flag is now using `--blocks-pruning` which starts from polkadot version 0.928
```
BulatSaif, line 111:
```suggestion
- The pruning flag is now using `--blocks-pruning` which starts from polkadot version v0.9.28
```

charts/node/README.md.gotmpl
```diff
@@ -186,6 +189,7 @@ An error will be thrown if any of those flags are set directly.
 - `--light`
 - `--database`
 - `--pruning`
+- `--blocks-pruning`
```
BulatSaif, line 192: This is change-log from  v3.x.x to v4.0.0 it should not be edited, 
```suggestion
```

charts/node/values.yaml
```diff
@@ -181,7 +181,8 @@ node:
   chainData:
     # -- Database backend engine to use
     database: rocksdb
-    # -- Set the amount of blocks to retain. If set to 0 archive node will be run
+    # -- Set the amount of blocks to retain.
+    # If set to 0 archive node will be run. If --pruning flags is used, set this to false.
```
BulatSaif, line 185:
```suggestion
    # If set to 0 archive node will be run. If deprecated `--pruning` flags is used in `node.flags`, set this to `false`.
```

charts/node/values.yaml
```diff
@@ -274,7 +275,8 @@ node:
     chainData:
       # -- Database backend engine to use for the collator relay-chain database
       database: rocksdb
-      # -- Set the amount of blocks to retain for the collator relay-chain database. If set to 0 archive node will be run
+      # -- Set the amount of blocks to retain for the collator relay-chain database. 
+      # If set to 0 archive node will be run. If --pruning flags is used, set this to false.
```
BulatSaif, line 279:
```suggestion
      # -- Set the amount of blocks to retain for the collator relay-chain database.
      # If set to 0 archive node will be run. If deprecated `--pruning` flags is used in `node.collatorRelayChain.flags`, set this to `false`.
```

charts/node/README.md.gotmpl
```diff
@@ -107,6 +107,9 @@ node:
 ```
 
 ## Upgrade
+### From v5.x.x to v5.5.0 (⚠️ breaking changes)
+- The pruning flag is now using `--blocks-pruning` which starts from polkadot version 0.928
+- The flag `--pruning` is now allowed in both .Values.node.flags and .Values.node.collatorRelayChain.flags. When using `--pruning`, ensure that the values of .Values.node.chainData.pruning and .Values.node.collatorRelayChain.chainData.pruning are explicitly set to false to maintain previous behavior.
```
BulatSaif, line 112:
```suggestion
- The flag `--pruning` is now allowed in both .Values.node.flags and .Values.node.collatorRelayChain.flags. When using `--pruning`, ensure that the values of .Values.node.chainData.pruning and .Values.node.collatorRelayChain.chainData.pruning are explicitly set to `false` to maintain previous behavior.
```

BulatSaif (approved):
Please apply README.md fixes before merging.

PR #319, opened 2024-01-02
set the default chainPath for common chains (kusama, westend, rococo)

PierreBesson:


charts/node/README.md.gotmpl
```diff
@@ -31,35 +31,6 @@ helm install polkadot-node parity/node
 
 This will deploy a single Polkadot node with the default configuration.
 
-### Public snapshots
```
PierreBesson, line 34: Removing this section that will be added to umbrella chart.

PR #326, opened 2024-01-23
Add genereric command/params for bridge-relay

PierreBesson:
Breaking change, revamp how the bridge relayer configuration is set.

### DIFF
--- a/charts/bridges-common-relay/templates/deployment.yaml
+++ b/charts/bridges-common-relay/templates/deployment.yaml
@@ -33,25 +33,8 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
-            {{- if .Values.relayHeaders.enabled }}
-            - "relay-headers"
-            - "{{ .Values.relayHeaders.name }}"
-            {{- range $key, $value := .Values.relayHeaders.params }}
-            - "--{{ $key }} {{ $value }}"
-            {{- end }}
-            {{- end }}
-            {{- if .Values.relayHeadersAndMessages.enabled }}
-            - "relay-headers-and-messages"
-            - "{{ .Values.relayHeadersAndMessages.name }}"
-            {{- range $key, $value := .Values.relayHeadersAndMessages.params }}
-            {{- if kindIs "slice" $value }}
-            {{- range $param := $value }}
-            - "--{{ $key }} {{ $param  }}"
-            {{- end }}
-            {{- else }}
-            - "--{{ $key }} {{ $value }}"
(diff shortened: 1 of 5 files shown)


charts/bridges-common-relay/templates/deployment.yaml
```diff
@@ -33,6 +33,20 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
+            {{- if .Values.command }}
```
BulatSaif, line 36: Kubernetes has separate field for `command`, if we set command in args it will confuse users.  
```
spec:
  containers:
  - name: command-demo-container
    command: ["printenv"]
    args: ["HOSTNAME", "KUBERNETES_PORT"]
```
currently the default  value for `command` is `ENTRYPOINT ["/home/user/bridge-entrypoint.sh"]`

PierreBesson: I'm not sure how to call it. But we have multiple commands within the relayer binary. Maybe we can call it `subcommand` ?

PierreBesson:
> we already have `Values.extraArgs` to put any genereric command.

Yes but with these variables it makes it more clear what needs to be set...

charts/bridges-common-relay/templates/deployment.yaml
BulatSaif: Maybe let's use a list instead of strings/maps.

I see at least 2 use cases not covered by this PR:
1. What will we do if we have 2 or 3 subcommands?
2. How do we pass the list of valid bash parameters:
```
--lane 00000001 \
--lane 00000002 \
--dry-run
```
I assume it would be like this:
```yaml
.Values.params:
  lane: '00000001'
  lane: '00000002'
  dry-run: ''
```
but this would result in invalid YAML (key `lane` present twice) and render output:
```
--lane=00000002 \
--dry-run=
```

I propose having a single `.Values.args: [ ]`.
For example, for code like this:
```
subcommand1  \
subcommand2 \
subcommand3 \
--lane 00000001 \
--lane 00000002 \
--dry-run
```
we would have values:
```yaml
.Values.args:
 - subcommand1
 - subcommand2
 - subcommand3
 - "--lane 00000001"
 - "--lane 00000002"
 - "--dry-run"
```

PierreBesson: OK let's go with your solution of only one "params" variable.

PierreBesson: I implemented the single var solution. Also removed the previous way to set flags so this is a breaking change.

BulatSaif (approved):

PR #333, opened 2024-03-18
[node] add tpl to init containers

BulatSaif:
There are situations when it is useful to run a node with a custom chainspec. In the current approach, you need to generate the chainspec, upload it, and set `CustomChainspecUrl`, or you need a custom binary. An alternative approach is to use an init container to generate and store the chain-spec in `/chain-data/chainspec.json`. In this PR, I added some changes which may be useful for doing so. It is now possible to mount a custom chainspec as a configmap as well.

Changes:
1. added tpl to  containers and initcontainers
2. added extra ports to service
3. added option to use custom chainspec file.


lazam (approved):

PierreBesson:
Could you add a test of an example custom chainspec in the `ci/` values file ?

BulatSaif:
> Could you add a test of an example custom chainspec in the `ci/` values file ?

I added `examples/local-rococo`

PierreBesson (approved):

PR #342, opened 2024-05-14
Publish the P2P port even if the pod is not ready

BulatSaif:
When `enableSidecarReadinessProbe` is set to `true`, the pod will not be marked as ready until it has enough peers and is synced. However, we need to open the P2P port for incoming connections to have more peers and sync faster.


charts/node/values.yaml
```diff
@@ -492,6 +492,8 @@ node:
       externalTrafficPolicy: Cluster
       # -- Service type
       type: NodePort # or ClusterIP or LoadBalancer
+      # -- Publish the P2P port even if the pod is not ready (e.g., node is syncing). It's recommended to keep this to true.
+      publishNotReadyAddresses: true
```
PierreBesson, line 496: Replace `NotReady` by `Unready`

PierreBesson (approved):
Approved but would prefer to rewrite NotReady to Unready.

charts/node/values.yaml
BulatSaif: Updated

PR #354, opened 2024-08-21
[node] fix warning

BulatSaif:
**Fix for warning during node deployment:**

```
coalesce.go:286: warning: cannot overwrite table with non-table for node.node.keys (map[])
```

**Solution:** Ensure that `node.node.keys` is defined consistently as a list in  values files. Check and update  values files to avoid conflicting data types for `node.node.keys`.


lazam (approved):

lazam:
Don't forget to bump the version

PR #362, opened 2025-01-09
[common] add template for ingress

BulatSaif:
It is useful to template ingress rules so that, between environments, only the host changes while the backend configuration remains the same.

Changes:
1. Add template to the ingress rule.
2. Remove invalid default value for TLS.
3. Bump Helm chart version.

### DIFF
--- a/charts/common/values.yaml
+++ b/charts/common/values.yaml
@@ -135,25 +135,27 @@ ingress:
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  # -- A list of hosts for the Ingress
+  # -- Hostname used for the default ingress rule. If .Values.ingress.rules is set, this host is not used.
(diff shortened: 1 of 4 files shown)


charts/common/values.yaml
```diff
@@ -135,25 +135,27 @@ ingress:
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  # -- A list of hosts for the Ingress
+  # -- Hostname used for the default ingress rule. If .Values.ingress.rules is set, this host is not used.
+  host: chart-example.local
```
ArshamTeymouri, line 139: would this also work if a service needs more than one domain configured?

BulatSaif: For any custom use case, the `.Values.ingress.rules` need to be updated. The default rule for most common use cases: forward all traffic from one domain to the HTTP port.

PR #376, opened 2026-03-03
node: add seed-per-node support

nprt:
Adds support for individual seeds per node.

The default usage is specified by `.Values.node.vault.keys.vaultKeyAppendPodIndex`, which defaults to `false`.

When enabled, the flag will index secrets per node's number (e.g. `aura-0`, `aura-1`, etc.).

The changes are backwards compatible, and aren't impacting the existing versions.


eduardspa (commented):
There's a stray "j" here. Otherwise, I think this is what we need to update the chart.

eduardspa (approved):

lexnv (approved):

